PR #82, opened 2023-05-23
Fix/js bundle

hectorfhurtado:
#### What?

- An attempt of removing all unused JS files along with the last trail of jQuery
- Lodash is now tree shaked.
- There are also fixes to some JS errors when testing these changes.


hectorfhurtado:
Before:

image

After:

image

PR #92, opened 2023-09-06
Style blog page and post page

aleksandraworhacz:
* Style blog posts page
* Style single post page


templates/pages/blog.html
```diff
@@ -15,10 +15,11 @@
 {{/partial}}
 
 {{#partial "page"}}
-    {{> components/common/breadcrumbs breadcrumbs=breadcrumbs}}
-    <section>
+    <section class="max-w-4xl px-2 md:px-4 mx-auto my-8">
```
hectorfhurtado, line 18: @aleksandraworhacz instead of `max-w-4xl` should it be `max-w-desktop` as with other pages?

templates/components/common/share.html
```diff
@@ -1,24 +1,22 @@
 {{#if settings.add_this.buttons}}
     {{assignVar 'encodedUrl' (encodeURI url)}}
     {{assignVar 'encodedTitle' (encodeURI head.title)}}
-    <ul class="o-list-bare flex gap-4 items-center mt-16">
+    <ul class="flex items-center mt-16 -mx-2 text-xl">
         {{#each settings.add_this.buttons}}
-            <li>
+            <li class="">
```
hectorfhurtado: @aleksandraworhacz if this is empty, we can remove it

templates/pages/blog.html
aleksandraworhacz: in cornerstone theme, the blog page has more narrow content than other pages. That's why I used `max-w-4xl` instead of `max-w-desktop` 

https://cornerstone-light-demo.mybigcommerce.com/journal/

hectorfhurtado: @aleksandraworhacz true, we could move the text content to the right in order to use that spacing? I ask because this would be the only place where we'd use a different max width.

PR #94, opened 2023-12-25
XCORNER-4 basic site setup and styling

mgbelegu:
#### Changes:

- Button styling
- Form fields styling
- Radio and checkbox styling
- Custom font


assets/scss/components/_button.scss
```diff
@@ -1,47 +1,144 @@
...
+    &:focus {
+        background: $color;
+        box-shadow: 0px 0px 0px 4px $shadow-color;
+        color: $color-white;
+    }
+}
+
+@mixin button_disabled() {
+    &:disabled {
+        opacity: 0.48;
```
aleksandraworhacz, line 23: Maybe set `cursor` as well ?

assets/scss/components/_button.scss
```diff
@@ -1,47 +1,144 @@
...
+    }
+}
+
+@mixin button_tertiary_base() {
+    padding: 12px 0;
+    background: $color-white;
+    text-align: center;
+    font-family: $font-sans-serif;
+    font-size: 16px;
```
aleksandraworhacz: How about using `rem` instead of `px` throughout the site?

assets/scss/components/_button.scss
```diff
@@ -1,47 +1,144 @@
...
+    @include button_focus($color-primary-400, $color-primary-shadow);
+    color: $color-primary-300;
+
+    &:hover {
+        background: $color-primary-400;
+        color: $color-white;
+    }
+}
+
+.c-button--primary-red {
```
aleksandraworhacz: I'm not a fan of name `secondary-red`.  Maybe `.c-button--warning` or `c-button--quaternary` or even `c-button--red` would fit better?

assets/scss/components/_button.scss
```diff
@@ -1,47 +1,144 @@
...
     &:focus {
-        background-color: $color-medium-grey;
-        border-color: $color-medium-grey;
-        color: $color-white;
+        color: $color-red-400;
     }
 }
+
+.button--large {
+    padding: 16px 24px;
```
aleksandraworhacz: `rem` so the site would scale nicer?

assets/scss/components/_button.scss
```diff
@@ -1,47 +1,144 @@
...
     text-align: center;
+    font-family: Assistant, $font-sans-serif;
+    font-size: 16px;
+    font-style: normal;
+    font-weight: 600;
+    line-height: 1.5;
+    outline: 0;
+    border: 0;
+    padding: 12px 16px;
+    cursor: pointer;
```
aleksandraworhacz, line 57: Consider using a unified order of properties. It could be alphabetical or grouped by type, that way the code will look more organized.

assets/scss/components/_form.scss
```diff
@@ -1,3 +1,7 @@
+/* SVG icon url's */
+$checkmark-icon: 'data:image/svg+xml;base64,CjxzdmcgeG1sbnM9Imh0dHA6Ly93d3cudzMub3JnLzIwMDAvc3ZnIiB3aWR0aD0iMTIiIGhlaWdodD0iMTAiIHZpZXdCb3g9IjAgMCAxMiAxMCIgZmlsbD0ibm9uZSI+CiAgPHBhdGggZD0iTTExIDFMNCA4TDEgNSIgc3Ryb2tlPSJ3aGl0ZSIgc3Ryb2tlLXdpZHRoPSIyIi8+Cjwvc3ZnPgo=';
+$dropdown-icon: 'data:image/svg+xml;base64,CjxzdmcgeG1sbnM9Imh0dHA6Ly93d3cudzMub3JnLzIwMDAvc3ZnIiB3aWR0aD0iMjQiIGhlaWdodD0iMjQiIHZpZXdCb3g9IjAgMCAyNCAyNCIgZmlsbD0ibm9uZSI+CiAgPHBhdGggZD0iTTYgOUwxMiAxNUwxOCA5IiBzdHJva2U9ImJsYWNrIiBzdHJva2Utd2lkdGg9IjEuNSIgc3Ryb2tlLWxpbmVjYXA9InJvdW5kIiBzdHJva2UtbGluZWpvaW49InJvdW5kIi8+Cjwvc3ZnPgo=';
```
aleksandraworhacz: Can't you use a path to local svg file?

assets/scss/components/_form.scss
```diff
@@ -21,3 +25,230 @@
...
+                    &::before {
+                        box-shadow: 0px 0px 0px 4px rgba(90, 100, 193, 0.24);
+                    }
+                }
+            }
+        }
+    }
+}
+
+.js-form-field-checkbox {
```
aleksandraworhacz: Classes with `.js-(...)` seem more suitable for linking js actions than styling

assets/scss/components/_form.scss
```diff
@@ -21,3 +25,230 @@
...
+        &:focus {
+            &::before {
+                box-shadow: 0px 0px 0px 4px rgba(90, 100, 193, 0.24);
+                transition: all 0.2s ease-in-out;
+            }
+        }
+    }
+
+    .js-input[type='checkbox'] {
+        &:checked {
```
aleksandraworhacz: this could be moved to line 189 and prevent duplicating selector `.js-input[type='checkbox']`

assets/scss/components/_form.scss
mgbelegu: There are instances where the styling of an error message in the form depends on the element with class with '.js-form-field-error' which is the class that is added to that element via JavaScript when an error occurs. Without this class if we apply error styling to that element, it will be shown even if the message is not an error message. I think we are good to use these classes as well and we shouldn't be constrained. Otherwise we will spend lots of time adding other classes and modifying most of the structure again, which is not something we should focus on this project. We should just focus on styling it to make it match the designs. But if you want maybe we should discuss with either Pawel or Marcin (if he is available) to see how to approach this

mgbelegu:

assets/scss/components/_button.scss
```diff
@@ -1,47 +1,146 @@
...
-    vertical-align: middle;
 
-    &:hover,
+    outline: 0;
+    border: 0;
+    padding: 0.75rem 1rem;
+    cursor: pointer;
+
+    @include button_transition();
+    @include button_disabled();
```
aleksandraworhacz: In general, it's good to have any `@include` at the top, so they won't overwrite your styling

assets/scss/components/_button.scss
```diff
@@ -1,47 +1,146 @@
...
-        color: $color-white;
+        color: $color-red-400;
     }
 }
+
+.button--large {
+    padding: 1rem 1.5rem;
+}
+
+.button--large-tertiary {
```
aleksandraworhacz: Instead of `--large-tertiary` maybe use `--large-vertically`?

assets/scss/components/_form.scss
```diff
@@ -21,3 +25,236 @@
...
+        box-shadow: 0px 0px 0px 4px rgba(90, 100, 193, 0.24);
+        transition: all 0.2s ease-in-out;
+    }
+}
+
+label {
+    color: $color-black;
+
+    font-family: $font-sans-serif;
+    font-size: 16px;
```
aleksandraworhacz: Could be `rem` :)

assets/scss/components/_form.scss
```diff
@@ -21,3 +25,236 @@
         }
     }
 }
+
+input[type='text'],
+input[type='email'],
+input[type='password'],
+textarea {
```
aleksandraworhacz: Consider using css classes instead of styling elements directly. Also if you want to keep it that way, this code should be moved to `elements` folder instead or be wrapped with .c-form if the styling should be limited only to the given component

assets/scss/components/_form.scss
```diff
@@ -21,3 +25,236 @@
...
+        border: 1px solid $color-primary-300;
+        transition: all 0.2s ease-in-out;
+    }
+
+    &:focus {
+        box-shadow: 0px 0px 0px 4px rgba(90, 100, 193, 0.24);
+    }
+}
+
+select {
```
aleksandraworhacz: Similar here, it should be either moved to `elements` or replaced with a proper CSS class like `.c-form__select`

assets/scss/components/_form.scss
```diff
@@ -21,3 +25,236 @@
...
+    gap: 8px;
+    padding: 16px;
+
+    border-radius: 4px;
+    border: 1px solid $color-neutrals-75;
+
+    background: $color-white;
+    color: $color-neutrals-500;
+    font-family: $font-sans-serif;
+    font-size: 16px;
```
aleksandraworhacz: Could be replaced with `rem` units

assets/scss/components/_form.scss
```diff
@@ -21,3 +25,236 @@
...
+        transition: all 0.2s ease-in-out;
+    }
+
+    &:focus {
+        box-shadow: 0px 0px 0px 4px rgba(90, 100, 193, 0.24);
+        transition: all 0.2s ease-in-out;
+    }
+}
+
+label {
```
aleksandraworhacz: similar situation as with `select`  (line 62)

assets/scss/components/_form.scss
```diff
@@ -21,3 +25,236 @@
...
+
+    margin-bottom: 8px;
+
+    small {
+        color: $color-red-300;
+        font-weight: 400;
+    }
+}
+
+.js-form-field-error {
```
aleksandraworhacz: Please try not to use `.js-` classes for styling. Those should be used only to handle javascript code.

assets/scss/components/_form.scss
```diff
@@ -21,3 +25,236 @@
...
+            border-radius: 50%;
+            background: $color-white;
+            border: 1px solid $color-neutrals-75;
+            transition: all 0.2s ease-in-out;
+        }
+
+        &:hover {
+            &::before {
+                border: 1px solid $color-primary-300;
+                transition: all 0.2s ease-in-out;
```
aleksandraworhacz: this could be extracted to a variable as you are using the same property in a few places
e.g. `$input-transition: all 0.2s ease-in-out`

assets/scss/components/_form.scss
```diff
@@ -21,3 +25,236 @@
...
+
+    small {
+        color: $color-red-300;
+        font-weight: 400;
+    }
+}
+
+.js-form-field-error {
+    label,
+    .js-inline-message {
```
aleksandraworhacz: Please try not to use `.js-` classes for styling. Those should be used only to handle javascript code.

assets/scss/components/_form.scss
```diff
@@ -21,3 +25,236 @@
...
+        color: $color-red-300;
+    }
+}
+
+.js-inline-message {
```
aleksandraworhacz: Please try not to use `.js-` classes for styling. Those should be used only to handle javascript code.

assets/scss/components/_form.scss
```diff
@@ -21,3 +25,236 @@
...
+    font-family: $font-sans-serif;
+    font-size: 16px;
+    font-style: normal;
+    font-weight: 400;
+    line-height: 1.5;
+}
+
+.js-form-field-radio {
```
aleksandraworhacz: Please try not to use `.js-` classes for styling. Those should be used only to handle javascript code.

assets/scss/components/_form.scss
```diff
@@ -21,3 +25,236 @@
...
+    .js-input[type='radio'] {
```
aleksandraworhacz: Please try not to use `.js-` classes for styling. Those should be used only to handle javascript code.

assets/scss/components/_form.scss
```diff
@@ -21,3 +25,236 @@
...
+
+        &:focus {
+            &::before {
+                box-shadow: 0px 0px 0px 4px rgba(90, 100, 193, 0.24);
+                transition: all 0.2s ease-in-out;
+            }
+        }
+    }
+
+    .js-input[type='radio'] {
```
aleksandraworhacz: Please try not to use `.js-` classes for styling. Those should be used only to handle javascript code.

assets/scss/components/_form.scss
```diff
@@ -21,3 +25,236 @@
...
+        &:hover {
+            &::before {
+                border-color: $color-primary-400;
+                transition: all 0.2s ease-in-out;
+            }
+        }
+
+        &:focus {
+            &::before {
+                box-shadow: 0px 0px 0px 4px rgba(90, 100, 193, 0.24);
```
aleksandraworhacz: Box shadow color could be stored as a scss variable

assets/scss/components/_form.scss
aleksandraworhacz: If there's some js logic that applies classes and determines elements display then it's all good. 
But here, you are mixing logic. Styling elements by js classes, when it's possible to add a dedicated CSS class (I'm not 100% sure if it's a case here) in my opinion makes code less readable and breaks the BEM convention. 

Note that all of my comments are suggestions and if you see that something makes no sense we can skip it or as you mentioned discuss with Paweł or Marcin.

aleksandraworhacz (commented):
I've added a bunch of comments like "please try not to use js classes for styling"  but if it's not possible to use css classes or it will take too much time/effort it's up to Paweł and Marcin.

assets/scss/components/_form.scss
mgbelegu: It makes sense. I will see which ones can be edited and I will add different classnames to seperate logic from js class names. Thanks Ola

mgbelegu:
I managed to add all classes where I can so we can use those for CSS and not use the JS classes and I've also tackled the px to rem conversions. Please review again. @aleksandraworhacz

assets/scss/components/_form.scss
```diff
@@ -21,3 +26,236 @@
         }
     }
 }
+
+.c-text-input,
+.c-password-input,
+.c-textarea-input {
```
aleksandraworhacz: I don't want to be pain in the ass, but since the file called `components/form` I'd assume the CSS classes should start with `.c-form...` :D so instead of `c-text-input` it would be better to have `c-form__input` with `c-form__input--text` for text specific styling

assets/scss/components/_form.scss
```diff
@@ -21,3 +26,236 @@
...
+        border: 0.063rem solid $color-primary-300;
+        transition: all 0.2s ease-in-out;
+    }
+
+    &:focus {
+        box-shadow: $form-field-box-shadow;
+    }
+}
+
+.c-select-input {
```
aleksandraworhacz: same here `c-select-input` suggests that there should be a file `components/select-input.scss`

Consider either moving input styling to `components/input.scss` or rename properly :) Also you could move all inputs' common styles to `c-form__input`/`c-input` (depending on which approach you choose) 

Seems like this could be considered a common style - at least at first glance
```
    display: flex;
    align-items: center;
    gap: 0.5rem;
    padding: 1rem;

    border-radius: 0.25rem;
    border: 0.063rem solid $color-neutrals-75;

    background: $color-white;
    color: $color-neutrals-500;
    font-family: $font-sans-serif;
    font-size: 1rem;
    font-style: normal;
    font-weight: 400;
    line-height: 1.5;

     transition: all 0.2s ease-in-out;
```

Also `transition: all 0.2s ease-in-out;` could be replaced with something like `$input-transition` variable. You are using the same transition in 11 places just in this file

mgbelegu:
@aleksandraworhacz Hi, I decided to keep all input styles inside "form.scss". Please review again and let me know if it looks good

assets/scss/components/_form.scss
```diff
@@ -21,3 +27,236 @@
         }
     }
 }
+
+.c-form__text-input,
+.c-form__password-input,
+.c-form__textarea-input {
```
aleksandraworhacz: Just one more small suggestion. I'd consider using `.c-form__input .c-form__input--text` instead. Then you could extract shared styling into `.c-form__input` class and keep the code DRY :)

aleksandraworhacz:
Added one small suggestion. Other than that approved :)

assets/scss/components/_form.scss
```diff
@@ -21,3 +27,209 @@
         }
     }
 }
+
+.c-form__input,
+.c-form__input,
+.c-form__input {
```
aleksandraworhacz, line 57: You don't need the same selector three times :D

assets/scss/components/_form.scss
```diff
@@ -21,3 +27,209 @@
...
+    display: block;
+    margin-bottom: 0.5rem;
+
+    small {
+        color: $color-red-300;
+        font-weight: 400;
+    }
+}
+
+.c-form__radio-field {
```
aleksandraworhacz: Do we need this class? It doesn't contain any self-styling and is used just to build more complex selectors - which is not the point of ITCS

assets/scss/components/_form.scss
```diff
@@ -21,3 +27,209 @@
...
+    .c-form__radio-input {
```
aleksandraworhacz: Duplicated class at line 93 and 130

assets/scss/components/_form.scss
```diff
@@ -21,3 +27,209 @@
...
+                    &::before {
+                        box-shadow: $form-field-box-shadow;
+                    }
+                }
+            }
+        }
+    }
+}
+
+.c-form__checkbox-field {
```
aleksandraworhacz: This also feels like it's not really needed (no styling for the class itself, just an extra wrapper)

aleksandraworhacz: Also it feels like `.c-form__checkbox-field` and `.c-form__radio-field` have a lot in common, maybe their code could be simplified and share common styles via `.c-form__choice-input` or something similar?

mgbelegu:
@aleksandraworhacz Changes made. Please review again :)

assets/scss/components/_form.scss
```diff
@@ -1,12 +1,44 @@
+/* SVG icon url's */
+$checkmark-icon: './../img/icons/checkmark.svg';
+$dropdown-icon: './../img/icons/dropdown.svg';
+$form-field-box-shadow: 0 0 0 0.25rem rgba(90, 100, 193, 0.24);
+$input-transition: all 0.2s ease-in-out;
+
+@mixin pseudoElementHoverFocusStyles() {
```
aleksandraworhacz, line 7: I'm not sure why this couldn't be a class but let's be it :D

aleksandraworhacz:
Approve, but I couldn't find a button to click 🙈

PR #96, opened 2023-12-26
XCORNER-1 snackbar over navigation

mgbelegu:
#### Changes:

- Styled the default announcement banner in BigCommerce to match the designs


assets/scss/components/_header.scss
```diff
@@ -26,3 +23,36 @@
...
 }
+
+.c-header-announcement-banner {
+    display: flex;
+    justify-content: center;
+    align-items: center;
+    background-color: $color-black;
+    padding: 16px 24px;
+
+    p {
```
aleksandraworhacz, line 34: please use css classes instead of styling directly element

assets/scss/components/_header.scss
```diff
@@ -26,3 +23,36 @@
...
+        font-weight: 400;
+        line-height: 1.5;
+        margin: 0;
+    }
+
+    strong {
+        font-weight: 600;
+    }
+
+    a {
```
aleksandraworhacz, line 49: Similar here, if css class can be applied, avoid nesting selectors and styling directly an element

PR #100, opened 2024-01-17
Xcorner 14 sign in page

mgbelegu:
#### Changes:

- Styling of sign in page


assets/scss/pages/contact-page.scss
```diff
@@ -0,0 +1,109 @@
...
+    margin: 0;
+    color: $color-black;
+    font-family: $font-semi-bold;
+    font-size: 32px;
+    font-style: normal;
+    font-weight: 600;
+    line-height: 1.25;
+}
+
+.c-contact-form {
```
marcinkrzeminski: @mgbelegu it would be better to make a separate component here, unless it's strictly related to this page, than I would go with `p-contact__form`.

assets/scss/pages/contact-page.scss
```diff
@@ -0,0 +1,109 @@
...
+        row-gap: 24px;
+        column-gap: 24px;
+
+        @include bp(medium) {
+            row-gap: 48px;
+            column-gap: 24px;
+        }
+
+        @include bp(large) {
+            row-gap: 48px;
```
marcinkrzeminski: @mgbelegu duplicate of property

assets/scss/pages/contact-page.scss
```diff
@@ -0,0 +1,109 @@
...
+            }
+
+            .c-form__label {
+                display: block;
+                margin-bottom: 8px;
+            }
+        }
+    }
+
+    textarea.c-form__input {
```
marcinkrzeminski: @mgbelegu is adding `textarea` necessary here?

assets/scss/pages/contact-page.scss
```diff
@@ -0,0 +1,109 @@
...
+        }
+    }
+
+    .c-button {
+        width: 100%;
+
+        @include bp(medium) {
+            width: fit-content;
+        }
+    }
```
marcinkrzeminski: @mgbelegu I'd suggest moving this to a button component page and make a variation out of it like `c-button--full`.

assets/scss/pages/contact-page.scss
```diff
@@ -0,0 +1,109 @@
...
+
+    br {
+        display: none;
+    }
+}
```
marcinkrzeminski: @mgbelegu is this necessary here? Can we avoid outputting <br> in the template?

assets/scss/pages/contact-page.scss
```diff
@@ -0,0 +1,109 @@
...
+
+.c-contact-us__page {
```
marcinkrzeminski: @mgbelegu this should be part of the `p-contact`

assets/scss/pages/contact-page.scss
```diff
@@ -0,0 +1,109 @@
...
+
+    @include bp(medium) {
+        max-width: 48.875rem;
+    }
+
+    p {
+        &:last-of-type {
+            margin-bottom: 0;
+        }
+    }
```
marcinkrzeminski: @mgbelegu I'd suggest creating `c-wysiwyg` component and nest all of these like `p`, `strong`, etc. inside this selector. Similar to something like `prose` does in Tailwind.

assets/scss/pages/login-page.scss
```diff
@@ -0,0 +1,168 @@
...
+    gap: 1.5rem;
+    width: 100%;
+
+    @include bp(medium) {
+        width: auto;
+    }
+
+    .c-form__field {
+        @include bp(medium) {
+            min-width: 21.37rem;
```
marcinkrzeminski, line 35: @mgbelegu please avoid these "magic values". I think nobody would notice if you would make it `21rem` or `21.5rem`

assets/scss/pages/login-page.scss
```diff
@@ -0,0 +1,168 @@
...
+        aspect-ratio: 1;
+        background-color: rgba(90, 100, 193, 0.08);
+        z-index: 0;
+        border-radius: 50%;
+    }
+}
+
+.required {
+    color: $color-red-300;
+}
```
marcinkrzeminski: @mgbelegu please make a separate utility class in a separate file like `u-color-red-300`. Also move it to `utils/color.scss` or something similar.

assets/scss/components/_breadcrumbs.scss
```diff
@@ -1,15 +1,35 @@
...
+    color: $color-primary-300;
+    font-family: $font-semi-bold;
+    font-size: 16px;
+    font-style: normal;
+    font-weight: 600;
+    line-height: 1.5;
+    text-decoration-line: underline;
+}
+
+.c-breadcrumbs__active {
```
marcinkrzeminski: @mgbelegu this is a great candidate for a modifier instead of an element. I suggest `c-breadrcumbs--active` or `c-breadcrumbs__link--active`

assets/scss/pages/contact-page.scss
```diff
@@ -0,0 +1,109 @@
+.c-contact-page__section {
```
marcinkrzeminski: @mgbelegu please use `p-` prefix for pages. Also, I think it would be better to name the main block like `p-contact` and then `p-contant__title` and so on.

assets/scss/pages/login-page.scss
```diff
@@ -0,0 +1,168 @@
...
+    padding: 1.5rem;
+    display: flex;
+    flex-direction: column;
+    gap: 1.5rem;
+
+    @include bp(medium) {
+        padding: 3rem 5rem;
+        gap: 3rem;
+    }
+}
```
marcinkrzeminski: @mgbelegu same as contact page. I'd suggest `p-login` as a main element on the page. Adjust children elements accordingly.

marcinkrzeminski (changes requested):
@mgbelegu please use proper image format for the main image. In this case, JPG is much more appropriate. PNG only if you really have to (like transparency). Also make sure to optimize it properly using ImageOptim or any online optimization tool like https://shortpixel.com/online-image-compression

assets/scss/pages/contact-page.scss
mgbelegu: i think it comes from recaptcha markup `{{{forms.contact.recaptcha.markup}}}`

mgbelegu:
@marcinkrzeminski Hi, can you please review again?

PR #104, opened 2024-01-27
XCORNER-17 default page styling

mgbelegu:
Default page styling


assets/scss/pages/page.scss
```diff
@@ -0,0 +1,34 @@
+.c-page-container {
```
marcinkrzeminski: @mgbelegu please use `p-` prefix for pages. I suggest we use `p-common` root class name so the class names would be `p-common__container` and `p-common__title`.

marcinkrzeminski: @mgbelegu if you want to keep like `c-page-container` and `c-page-title` this should be separate files with the same names (excluding the prefix). Decide yourself which approach you feel is better 👍🏻.

marcinkrzeminski (changes requested):

PR #105, opened 2024-02-07
Xcorner 12 category and product listing page

mgbelegu:


mgbelegu:
@marcinkrzeminski Hi, can you please review this one?

assets/scss/components/_grid.scss
```diff
@@ -0,0 +1,12 @@
+.c-grid__list {
```
marcinkrzeminski: @mgbelegu I think this one should be `c-grid-list` as the base. The one you created implies it's a child element of `c-grid`.

assets/scss/components/_paginator.scss
```diff
@@ -0,0 +1,38 @@
...
+
+.c-paginator__item {
+    margin: 0;
+}
+
+.c-paginator__link {
+    display: block;
+    width: 0.75rem;
+    height: 0.75rem;
+
```
marcinkrzeminski: @mgbelegu please remove that blank line

assets/scss/components/_sidebar.scss
```diff
@@ -0,0 +1,32 @@
+.c-sidebar__box {
```
marcinkrzeminski, line 7: @mgbelegu, I suggest using the sidebar-bar for the component/block here. So, the class structure would be: `c-sidebar-box`, `c-sidebar-box__list`, and so on.

assets/scss/components/_wishlist.scss
```diff
@@ -0,0 +1,13 @@
+.c-wishlist {
+    position: absolute;
+    bottom: 0;
+    right: 0;
+}
+
+.c-wishlist-add__button {
```
marcinkrzeminski: @mgbelegu another place where the BEM seems to be a bit mixed. I think it this case it should be `c-wishlist__add-button` or simply `c-wishlist__button` to make it less explicit. You can always use the modifier to give more context like `c-wishlist__button--primary`

assets/scss/pages/auth.scss
```diff
@@ -0,0 +1,22 @@
...
+}
+
+.p-auth__title {
+    color: $color-black;
+    font-family: $font-semi-bold;
+    font-size: 2.625rem;
+    font-style: normal;
+    font-weight: 600;
+    line-height: 1.33;
+
```
marcinkrzeminski: @mgbelegu please remove that blank line here.

assets/scss/pages/page.scss
```diff
@@ -0,0 +1,34 @@
+.p-page__container {
+    padding: 1.5rem;
+    display: flex;
+    flex-direction: column;
+    gap: 1.5rem;
+
+    @include bp(medium) {
+        padding: 3rem 5rem;
+        gap: 3rem;
+    }
```
marcinkrzeminski: @mgbelegu wouldn't it be better if we make a general container component for these?

assets/scss/pages/page.scss
```diff
@@ -0,0 +1,34 @@
...
+
+    p {
+        color: $color-neutrals-500;
+        font-family: $font-regular;
+        font-size: 1.125rem;
+        font-style: normal;
+        font-weight: 400;
+        line-height: 1.5rem;
+    }
```
marcinkrzeminski: @mgbelegu if possible move these to the `c-wysisyg` component.

marcinkrzeminski (changes requested):

assets/scss/pages/page.scss
mgbelegu: @marcinkrzeminski what do you mean by this? it's better to create a "div" element for example with class name .c-container that handles this styling?

assets/scss/components/_card.scss
```diff
@@ -0,0 +1,61 @@
...
+}
+
+.c-card__figure {
+    border-radius: 0.25rem;
+    margin-bottom: 0.75rem;
+    position: relative;
+    overflow: hidden;
+}
+
+.c-card__image {
```
marcinkrzeminski: @mgbelegu, don't you think that `c-card__image-wrapper` suits better here? And then, instead of nesting the ``, we could use `c-card__image` for the image itself.

assets/scss/components/_facet.scss
```diff
@@ -0,0 +1,58 @@
+.c-facets {
+    width: 100%;
+}
```
marcinkrzeminski: @mgbelegu this is a perfect candidate for a utility class like `u-width-full` or something similar.

assets/scss/components/_paginator.scss
```diff
@@ -0,0 +1,37 @@
+.c-paginator {
```
marcinkrzeminski: @mgbelegu I suggest renaming this component to `c-pagination`, which describes the section. The term `c-paginator` implies its function rather than being a section name.

templates/components/category/sort-box.html
```diff
@@ -1,36 +1,36 @@
...
-        </option>
-        {{#if result_count}}
-            <option value="relevance"{{#if sort '==' 'relevance'}} selected{{/if}}>
-                {{lang 'common.sorter.relevance'}}
-            </option>
-        {{/if}}
-    </select>
-</fieldset>
+<fieldset class="js-form-fieldset">
+    <label for="sort c-form__label">
```
marcinkrzeminski: @mgbelegu please move the selector to the class attribute ;).

templates/components/products/related.html
```diff
@@ -5,7 +5,7 @@
 <ul class="js-product-list" {{#if settings.data_tag_enabled}} data-list-name="Related Products" {{/if}}>
     {{#each this}}
         <li>
-            {{>components/products/card settings=../settings theme_settings=../theme_settings event="list" position=(add @index 1)}}
+            {{>components/products/card-alternate settings=../settings theme_settings=../theme_settings event="list" position=(add @index 1)}}
```
marcinkrzeminski, line 8: @mgbelegu this is nit-picking, but `card-alternative` looks better to my taste. Feel free to ignore this comment if you feel different 👍🏻.

marcinkrzeminski (changes requested):
Good job on that huge PR. Only a few adjustments.

templates/components/category/sort-box.html
mgbelegu: @marcinkrzeminski my bad 😅 great catch though

templates/components/products/related.html
mgbelegu: @marcinkrzeminski I use alternate, because that word is used by default in other bigcommerce areas thats why. Let's keep it this way if you don't mind :)

assets/scss/pages/page.scss
marcinkrzeminski: @mgbelegu yes exactly. Move out outside of the page.scss and make a separate component out of it.

marcinkrzeminski (approved):

PR #113, opened 2024-03-04
XCORNER-20 compare products page

mgbelegu:


assets/scss/pages/page.scss
```diff
@@ -13,11 +13,16 @@
...
-    line-height: 1.33;
     margin: 0;
+
+    @include bp(medium) {
+        font-size: rem-calc(42px);
+        line-height: rem-calc(56px);
+    }
 }
 
 .c-login__form {
```
aleksandraworhacz: woah, shouldn't it be in `components/login` ? :D

assets/scss/pages/page.scss
```diff
@@ -13,11 +13,16 @@
 .p-common__title {
```
aleksandraworhacz: Hm, I'm not sure if I like this approach. What's the reason for using new "pages" level instead of let's say "object"? In `assets/scss/theme.scss` you have 

```
@import 'settings/settings';
@import 'tools/tools';
@import 'generic/generic';
@import 'elements/elements';
@import 'vendor/vendor';
@import 'objects/objects';
@import 'components/components';
@import 'utilities/utilities';
@import 'widgets/widgets';
@import 'pages/pages';
```

Where the idea was to start with most generic and broader style scopes and narrow down the focus/scope to very specific elements/properties (therefore the inverted triangle in ITCSS). But now, after pretty narrow `utilities`, you are importing rather broad `widgets` and even more broad `pages`? 

I know it's not strictly related to this MR, but maybe it's worth to revisit ? :)

aleksandraworhacz (commented):
Added few comments :)

assets/scss/pages/page.scss
mgbelegu: probably issue when i was rebasing. i deleted that code because its duplicate :)

assets/scss/pages/page.scss
mgbelegu: @aleksandraworhacz let's keep this as is for now and keep it in mind later on. Marcin didn't tell me anything about this usage when I was working on it. Since we are almost at the end of the first demo I don't want to change this and spend more time with it. After the demo we can regroup see what we can improve from this first stage of development and get approval (because i think that if we change this now it will take more time to change)

also, let's say we work on this in the future. how would you do it? remove widgets and pages and then just create component scss files for every component on any page and widget? so basically just focusing on components and not seperating them by pages

aleksandraworhacz: > so basically just focusing on components and not seperating them by pages

Exactly. Focusing on pages/widgets instead of components will give us a headache when we end up with much of "look alike" code :/

aleksandraworhacz: Also, in this particular case (`assets/scss/pages/compare.scss`) I think it would be better to create "product-card-details" component (or maybe even conditional chunk of code inside the `card`) since it's a brand new file, and changing it won' t affect other components I think. We could postpone improving already exisitng code, but based on my experience, the "right time" almost never comes for this kind of change, and then everyone just forgets about it.

aleksandraworhacz:
Added few comments

mgbelegu:
@aleksandraworhacz Hi, I've made some updates regarding removing the "pages" stylesheet but I can continue with those later on in another task I created so we can make it to Wednesday's demo. Please review again

assets/scss/components/_product.scss
```diff
@@ -1,3 +1,178 @@
...
+    a {
+        color: $color-primary-300;
+    }
+}
+
+.c-product__pricing {
+    display: flex;
+    gap: rem-calc(8px);
+
+    .c-price__without-tax {
```
aleksandraworhacz: Few things I found not quite right here. 

1. Why can't we have `.c-product__pricing--no-tax` instead of `.c-price__without-tax` ? this `c-price` suggest that it's a separate component which I don't think is a case here
2. It's a bit unclear what `&--non-sale` does in this case. In general it's better to avoid negation in class naming as they may be confusing e.g. `--non-sale` means that the default price is "on sale"? 🤔  Definitelly better approach would be to use `--sale`, than everything is crystal clear :) 
3. I'm not sure why `--non-sale` relates to `.c-price__without-tax` instead of the parent: `.c-product__pricing`?

assets/scss/objects/_container.scss
```diff
@@ -0,0 +1,11 @@
+.o-container {
+    padding: 1.5rem;
```
aleksandraworhacz, line 2: When you create a container object, it's good to set `width` and `max-width` so the content won't get stretched to the viewport. But you don't need to do it in this case, as I'll merge my changes to the same file 😅

aleksandraworhacz (approved):
One teeny tiny comment added :) Otherwise approved :)

PR #114, opened 2024-03-06
Xcorner 21 cart page

mgbelegu:


assets/js/theme/custom/add-reviews-to-cart-items.js
```diff
@@ -0,0 +1,62 @@
...
+                .then((res) => res.json())
+                .then((data) => {
+                    const summationOfRatings = data?.data?.site?.product?.reviewSummary?.summationOfRatings
+                    const numberOfReviews = data?.data?.site?.product?.reviewSummary?.numberOfReviews
+                    const ratingElement = document.querySelector(`[data-type-product-id="${productId}"] .c-cart__item-rating`)
+                    const ratingLink = document.querySelector(`[data-type-product-id="${productId}"] .c-cart__item-rating-link`)
+
+                    ratingLink.innerHTML = `${numberOfReviews} reviews`
+
+                    const filledStarSVG = `
```
aleksandraworhacz: Not a fan of this solution as you hardcode the SVG inside js. Can't we set a ratings in element's data and based on this value display correct number of stars?

assets/scss/components/_cart.scss
```diff
@@ -0,0 +1,171 @@
...
+    height: auto;
+    max-width: 17.75rem;
+    object-fit: cover;
+
+    @include bp(small) {
+        width: 100%;
+    }
+}
+
+.c-cart__item-image {
```
aleksandraworhacz, line 42: Seems like `c-cart__item-image` could be moved to a separate component?

assets/scss/pages/cart.scss
```diff
@@ -0,0 +1,5 @@
+.p-cart__totals {
```
aleksandraworhacz: Why it's here? You have `components/totals` which seems like a perfect place for it

assets/scss/pages/page.scss
```diff
@@ -11,154 +11,12 @@
 }
 
 .p-common__title {
```
aleksandraworhacz: In this case I want you to leave it, because I've already worked on a typography. But in general, instead of creating `title` class inside `pages` - which doesn't make much sense, consider creating an object like `o-title` that could be reused in multiple places. It could have modifiers like `o-title--big` etc.

Try to think about a given part not through lenses of its location, but more like "what it does and how it could be used in different way".  I think in 90% of situations, when you want to put some class into `common` or `pages` objects ar your answers, especially if something if highly common :)

templates/components/cart/content.html
```diff
@@ -1,192 +1,135 @@
...
+                            {{#if can_modify}}
+                                <button 
+                                    class="js-cart-update c-cart__item-qty-btn"
+                                    data-cart-itemid="{{id}}"
+                                    data-action="dec"
+                                >
+                                    <span class="u-hidden-visually">
+                                        {{lang 'products.quantity_decrease' name=name}}
+                                    </span>
+                                    <svg xmlns="http://www.w3.org/2000/svg" width="24" height="24" viewBox="0 0 24 24" fill="none">
```
aleksandraworhacz: Please use SVG sprite https://github.com/xfiveco/xcorner?tab=readme-ov-file#icons instead of hardcoding icons

templates/components/cart/content.html
```diff
@@ -1,192 +1,135 @@
...
+                            {{#if can_modify}}
+                                <button 
+                                    class="js-cart-update c-cart__item-qty-btn"
+                                    data-cart-itemid="{{id}}"
+                                    data-action="inc"
+                                >
+                                    <span class="u-hidden-visually">
+                                        {{lang 'products.quantity_increase' name=name}}
+                                    </span>
+                                    <svg xmlns="http://www.w3.org/2000/svg" width="24" height="24" viewBox="0 0 24 24" fill="none">
```
aleksandraworhacz: SVG sprite

templates/components/cart/content.html
```diff
@@ -1,192 +1,135 @@
...
+                                --
+                            {{/or}}
+                            {{#or can_modify (if type '==' 'GiftCertificate')}}
+                                <button 
+                                    class="js-cart-remove c-cart__button-remove"
+                                    data-cart-itemid="{{id}}"
+                                    data-confirm-delete="{{lang 'cart.confirm_delete'}}"
+                                    aria-label="{{lang 'cart.remove_item' name=name}}"
+                                >
+                                    <svg xmlns="http://www.w3.org/2000/svg" width="24" height="24" viewBox="0 0 24 24" fill="none">
```
aleksandraworhacz: SVG Sprite

templates/components/cart/content.html
```diff
@@ -1,192 +1,135 @@
...
+                                    </span>
+                                    <svg xmlns="http://www.w3.org/2000/svg" width="24" height="24" viewBox="0 0 24 24" fill="none">
+                                        <path d="M8 12H16" stroke="#373D3F" stroke-width="1.5" stroke-linecap="round" stroke-linejoin="round"/>
+                                        <path d="M12 16V8" stroke="#373D3F" stroke-width="1.5" stroke-linecap="round" stroke-linejoin="round"/>
+                                        <path d="M9 22H15C20 22 22 20 22 15V9C22 4 20 2 15 2H9C4 2 2 4 2 9V15C2 20 4 22 9 22Z" stroke="#373D3F" stroke-width="1.5" stroke-linecap="round" stroke-linejoin="round"/>
+                                    </svg>
+                                </button>
+                            {{/if}}
+                        </div>
+                        <div class="c-cart__item-total">
```
aleksandraworhacz: could it be a separate component called `c-order-total` ? The same data is being displayed in order's details on account page so we could make the component more general

aleksandraworhacz (commented):
Added few comments

mgbelegu:
@aleksandraworhacz Hi, can you please review again? Regarding p-common__title, I've handled that in XCORNER-20, maybe I will merge that first and rebase this one

templates/components/cart/content.html
```diff
@@ -25,7 +25,42 @@
                             {{name}}
                         </a>
                         <div class="c-cart__item-rating-wrapper">
-                            <span class="c-cart__item-rating" role="img" aria-label=""></span>
+                            <div class="c-cart__item-rating" role="img" aria-label="">
+                                <span class="c-cart-item__filled-stars">
+                                    <svg class="u-hidden" xmlns="http://www.w3.org/2000/svg" width="12" height="12" viewBox="0 0 12 12" fill="none">
```
aleksandraworhacz: Everytime you repeat the same code, the red light should appear above your head saying "there must be a better way to approach this" :D 

Please create a separate component for the star, where you could pass the value "filled" or "unfilled"  and you could either use SVG sprite (https://github.com/xfiveco/xcorner?tab=readme-ov-file#icons) and toggle visibility similar as you do now. Or even better, you could pass the rating value into the "product rating" component and pass the rating then display star svg + clip path to present the exact rate value. 

Example for the single star: 
```
<svg>
  <defs>
    <clipPath id="star">
      <path d="...." />
    </clipPath>
  </defs>

  <rect style="width: var(--width, 100%); fill: var(--color, #dd2424)" clip-path="url(#star)"/>
</svg>
```

templates/pages/cart.html
```diff
@@ -23,7 +23,7 @@ <h1 class="p-common__title">{{lang 'cart.title'}}</h1>
             {{> components/cart/content}}
         </div>
 
-        <div class="js-cart-totals p-cart__totals">
+        <div class="js-cart-totals c-totals">
```
aleksandraworhacz: Ideally, it would be separated into a component named "totals" based on its name - especially if you are reusing in in multiple places. 

Also you already have `cart` component, so you could use "c-cart__totals" instead if it's more like a "one time" component

aleksandraworhacz (approved):
Added few small comments, otherwise approved :)

mgbelegu:

PR #115, opened 2024-03-07
Xcorner 29 modals

mgbelegu:


assets/js/theme/common/product-details.js
```diff
@@ -459,7 +459,7 @@ export default class ProductDetails extends ProductDetailsBase {
 
         // Add item to cart
         utils.api.cart.itemAdd(normalizeFormData(new FormData(form)), (err, response) => {
-            currencySelector(response.data.cart_id)
+            currencySelector(response?.data.cart_id)
```
aleksandraworhacz, line 466: I think we are missing the code for catching the error, something like

```
if (err || !response || response.data.error) {
      handle the error code
      throw new Error(err) or return to prevent executing the rest of code
}
```

assets/js/theme/custom/add-to-cart-modal.js
```diff
@@ -24,7 +24,7 @@ class AddToCartWithModal {
 
                 // Add item to cart
                 utils.api.cart.itemAdd(normalizeFormData(formData), (err, response) => {
-                    currencySelector(response.data.cart_id)
+                    currencySelector(response?.data.cart_id)
```
aleksandraworhacz, line 30: Similar here, I don't see handling the error. I know it's not a part of your  changes tho

assets/scss/components/_modal.scss
```diff
@@ -1,27 +1,91 @@
...
-    overflow-y: auto;
+    border: 0;
+    outline: none;
 }
 
-.c-modal.is-open > button {
-    position: absolute;
-    top: rem-calc(10px);
-    right: rem-calc(10px);
+.c-modal__header-title {
```
aleksandraworhacz: I think you overcomplicated the class name a little.  You could use `.c-modal__title`  - the simpler the better :)

assets/scss/components/_modal.scss
```diff
@@ -1,27 +1,91 @@
...
+    font-family: $font-semi-bold;
+    font-size: rem-calc(24px);
+    line-height: rem-calc(40px);
+    font-weight: 600;
+    letter-spacing: 0;
+    text-align: left;
+}
+
+.c-modal__button-wrapper {
```
aleksandraworhacz: It is not a big issue, but it would be good to keep classes that refer to the same element (button, button-wrapper etc.) near each other.

assets/scss/components/_modal.scss
```diff
@@ -1,27 +1,91 @@
...
+    display: flex;
+    gap: rem-calc(12px);
+    align-items: center;
+}
+
+.c-modal__product-name {
```
aleksandraworhacz: Hmm, seems like this class is very specific. If the modal is a separate component, that can contain information not necessarily strictly connected to a product, the class `c-modal__product-name` seems a bit too precise? Maybe we could use a more general name like `c-modal__heading` / `c-modal__subheading` / `c-modal__title` etc.

lang/en.json
```diff
@@ -134,7 +134,9 @@
             "order_subtotal": "Order subtotal",
             "continue_shopping": "Continue Shopping",
             "view_or_edit_cart": "View or edit your cart",
-            "you_might_also_like": "You may also like"
+            "you_might_also_like": "You may also like",
+            "woohoo": "Woohoo!",
+            "message": " was added to your cart."
```
aleksandraworhacz, line 140: You should also add those to other locales for the sake of keeping all of them (more or less) consistant

templates/components/common/modal/modal-close-btn.html
```diff
@@ -1,9 +1,11 @@
 <button 
-    class="js-modal-close"
+    class="js-modal-close c-modal__close-button"
     type="button"
     title="{{lang 'common.close'}}"
     {{additional_attr}}
 >
     <span class="u-hidden-visually">{{lang 'common.close'}}</span>
-    <span aria-hidden="true">&#215;</span>
+    <svg width="14" height="14" viewBox="0 0 14 14" fill="none" xmlns="http://www.w3.org/2000/svg">
```
aleksandraworhacz: Instead of hardcoding the icon, please use SVG sprite

aleksandraworhacz (commented):
Added few comments 🙈

mgbelegu:

aleksandraworhacz:
Looks good, approved :)

aleksandraworhacz (approved):

PR #116, opened 2024-03-12
XCORNER 26 Account Pages

aleksandraworhacz:
Added styling for account pages:
- login
- orders list
- order's details
- addresses
- edit address
- messages
- wishlist list
- wishlist's details
- settings

Additionally, introduced few improvements like:
- icon object
- container object
- grid object
- added text and heading objects


assets/js/theme/custom/order-product-data.js
```diff
@@ -0,0 +1,55 @@
+/* @TODO: Need to figure out a way to hide API TOKEN on the front. */
```
mgbelegu: @aleksandraworhacz The Storefront API token coming from BigCommerce is something that was designed to be available in the frontend. I don't think there is a way to hide it in the frontend.

assets/js/theme/custom/order-product-data.js
```diff
@@ -0,0 +1,55 @@
...
+export default function getOrderProductData(apiToken) {
+    const selectors = {
+        item: '.js-card-order',
+        url: '.js-dynamic-url',
+        wishlist: '.js-wishlist-add',
+    }
+
+    const items = Array.from(document.querySelectorAll(selectors.item))
+    items.forEach((item) => {
```
mgbelegu, line 9: Maybe check if items exist first as an extra precaution, unless you are sure they will exist 100% of the times

assets/js/theme/custom/order-product-data.js
```diff
@@ -0,0 +1,55 @@
...
+        fetch('/graphql', {
+            method: 'POST',
+            credentials: 'same-origin',
+            headers: {
+                'Content-Type': 'application/json',
+                Authorization: `Bearer ${apiToken}`,
+            },
+            body: JSON.stringify({
+                query: `
+                        query myQuery {
```
mgbelegu: Please rename `query myQuery` to a proper name for what the query is supposed to do

templates/components/account/order-contents.html
```diff
@@ -1,14 +1,20 @@
-<div>
+{{!-- 
+    @TODO: Remove asap.
```
mgbelegu: If the data doesn't come through handlebars or stencil, injecting data with JavaScript using the storefront API is usually how it's done in BigCommerce. I've encountered this issue with multiple other BC projects

mgbelegu (approved):
Great job with this big MR

mgbelegu:
@aleksandraworhacz Just two small comments regarding some changes, but approved overall

assets/js/theme/custom/order-product-data.js
aleksandraworhacz: Not an issue really. If there's no `selectors.item` we will end up with an empty array. And calling `forEach` on empty array won't throw an error :)

mgbelegu (approved):
approved :)

PR #118, opened 2024-03-14
Xcorner 30 search and product view fixes

mgbelegu:
- Fixed search popup
- Fixed product view product option styling


assets/icons/test.svg
aleksandraworhacz, line 1: Duplicate of "account.svg"?

assets/icons/search-black.svg
aleksandraworhacz, line 1: Remove "black" from the name

assets/icons/menu-close.svg
aleksandraworhacz, line 1: This could be used not only in menu, so maybe use "close" name?

assets/js/theme/common/product-details.js
```diff
@@ -109,7 +109,7 @@ export default class ProductDetails extends ProductDetailsBase {
             bannerUtils.dispatchProductBannerEvent(productAttributesData)
         }
 
-        $productOptionsElement.style.display = 'block'
+        $productOptionsElement.style.display = 'flex'
```
aleksandraworhacz, line 112: Is there a reason why this is handled in js instead CSS? At first sight it seems like CSS would be a much simpler solution :)

assets/js/theme/custom/mega-menu-search.js
```diff
@@ -0,0 +1,29 @@
+/**
+ * Open mega menu on search button click
+ */
+export default function megaMenuSearch() {
+    const MOBILE_WIDTH = 768
+    const megaMenu = document.querySelector('.c-navigation__megamenu')
```
aleksandraworhacz: please use `.js-...` classname for javascript operations :)

assets/scss/components/_menu.scss
```diff
@@ -36,6 +36,8 @@
 
 .c-menu__close {
     display: none;
+    max-width: rem-calc(16px) !important;
+    max-height: rem-calc(16px) !important;
```
aleksandraworhacz: I wonder, why we need those `!important`s here? When using ITCSS you shouldn't need to use import :)

assets/scss/utilities/_hide.scss
```diff
@@ -28,9 +28,18 @@
...
         display: none;
     }
 }
+
+
+.u-hidden\@mobile {
+    display: none;
+
+    @include bp(medium) {
+        display: flex;
```
aleksandraworhacz: `u-hidden@mobile` shouldn't overwrite styles when the element is visible. It will potentially break styling for elements that have different `display` property e.g. `inline` or `grid` 

Use `bp-down` instead

templates/components/products/options/swatch.html
```diff
@@ -59,19 +59,23 @@
                     <span title="{{this.label}}" style="background-color: #{{data.[0]}}"></span>
                     <span title="{{this.label}}" style="background-color: #{{data.[1]}}"></span>
                     <span title="{{this.label}}" style="background-color: #{{data.[2]}}"></span>
+                    <span>{{this.label}}</span>
```
aleksandraworhacz, line 64: Couldn't you use `for` loop and display as many labels as elements in `data` ? 🤔 
I know that the `if` wasn't created now, but seems like it could be simplified I think

templates/components/products/product-view.html
```diff
@@ -136,20 +136,21 @@ <h1 class="js-product-view-title c-product__title">
...
-                <div class="js-product-option-change" style="display:none;">
+                <div class="js-product-option-change c-product__options" style="display:none;">
                     {{inject 'showSwatchNames' theme_settings.show_product_swatch_names}}
                     {{#each product.options}}
                         {{{dynamicComponent 'components/products/options'}}}
                     {{/each}}
                 </div>
-                <div class="js-form-field js-form-field-stock{{#unless product.stock_level}} u-hidden-visually{{/unless}}">
+
+                {{!-- <div class="js-form-field js-form-field-stock{{#unless product.stock_level}} u-hidden-visually{{/unless}}">
```
aleksandraworhacz: Remove commented code? :D

assets/js/theme/common/product-details.js
mgbelegu: This was already implemented, i just wanted to use flex. But if you don't mind I can apply `!important` to my SCSS where I needed the 'flex' display. Let me know

templates/components/products/product-view.html
mgbelegu: okay, will remove since we don't show the stock in our current design

mgbelegu:
@aleksandraworhacz Can you please review again?

assets/icons/menu-close.svg
aleksandraworhacz, line 1: Coulndn't this icon be used in modal's close button? If so, maybe rename to "close"

assets/js/theme/common/product-details.js
aleksandraworhacz: If moving this to the stylesheet will require `!important` it would be good to revisit its styling. In such case it's ok to leave it as is for now.

assets/js/theme/custom/mega-menu-search.js
```diff
@@ -0,0 +1,29 @@
...
+    }
+
+    const updateEventListeners = () => {
+        const windowWidth = window.innerWidth
+
+        if (searchButton) {
+            if (windowWidth < MOBILE_WIDTH) {
+                searchButton.removeEventListener('click', openDesktopMenu)
+            } else {
+                searchButton.addEventListener('click', openDesktopMenu)
```
aleksandraworhacz, line 23: If you are calling `updateEventListeners` on `resize` wouldn't it results with adding multiple (the same) click events attached to searchButton?

assets/scss/components/_navigation.scss
```diff
@@ -90,6 +90,17 @@
     cursor: pointer;
 }
 
+.c-navigation__icon {
+    max-width: rem-calc(24px);
+    max-height: rem-calc(24px);
+    fill: transparent;
+
+    &.c-menu__close {
```
aleksandraworhacz: What's the reason for mixing two different components in one element? Can't we have `c-navigation__icon--close` ?

assets/scss/components/_quick-search.scss
```diff
@@ -4,19 +4,25 @@
...
+    justify-content: center;
+    align-items: center;
 
-    color: $color-neutrals-500;
-    font-family: $font-regular;
-    font-size: 1rem;
-    font-style: normal;
-    font-weight: 400;
-    line-height: 1.5;
+    .c-form__input {
```
aleksandraworhacz: Please don't nest selectors and don't mix them. If form needs to be styled differently for this particular case, just create modifier class `c-form__input--search`

assets/scss/utilities/_hide.scss
```diff
@@ -28,9 +28,15 @@
 }
 
 .u-hidden\@medium {
-    display: block;
+    display: flex;
```
aleksandraworhacz: If the classname is meant to hide elements, it should only add "display: none" when needed, without overwriting element's styles when it's visible

aleksandraworhacz (commented):
Comments added :)

assets/scss/utilities/_hide.scss
```diff
@@ -28,9 +28,13 @@
...
 
 .u-hidden\@medium {
-    display: block;
-
     @include bp(medium) {
         display: none;
     }
 }
+
+.u-hidden\@mobile {
```
aleksandraworhacz: It would be good to have consistent logic here. Now `u-hidden\@medium` and `u-hidden\@mobile` works different, as one of them hide elements on given breakpoint and up, and the other below. This will cause a headache in the future.

templates/components/products/options/swatch.html
aleksandraworhacz: up, seems like you've missed this one :)

assets/js/theme/custom/mega-menu-search.js
aleksandraworhacz: up

assets/scss/utilities/_hide.scss
```diff
@@ -19,18 +19,20 @@
     display: none !important;
 }
 
-.u-hidden\@small {
-    display: none;
+.u-hidden\@small-down {
```
aleksandraworhacz, line 22: Isn't it a duplication of `u-hidden\@medium-down` ? :D

mgbelegu:

assets/js/theme/custom/mega-menu-search.js
```diff
@@ -0,0 +1,37 @@
...
+        searchInputs?.forEach((input) => {
+            input.focus()
+        })
+    }
+
+    const updateEventListeners = () => {
+        const windowWidth = window.innerWidth
+        if (searchButton) {
+            if (hasAttachedEvent) {
+                searchButton.removeEventListener('click', openDesktopMenu)
```
aleksandraworhacz: What's the purpose of removing the "click" listener here?

assets/js/theme/global/mobile-menu-toggle.js
```diff
@@ -96,6 +100,16 @@ export class MobileMenuToggle {
         this.$header.classList.add('is-open')
         this.$scrollView.scrollIntoView(true)
 
+        window.scrollTo({
+            top: 0,
+        })
+
+        this.$searchInputs.forEach((input) => {
+            setTimeout(() => {
```
aleksandraworhacz: Have you tried using `requestAnimationFrame` instead `setTimeout` ?

aleksandraworhacz (commented):
Comments added

assets/js/theme/global/mobile-menu-toggle.js
```diff
@@ -96,6 +100,16 @@ export class MobileMenuToggle {
         this.$header.classList.add('is-open')
         this.$scrollView.scrollIntoView(true)
 
+        window.scrollTo({
+            top: 0,
+        })
+
+        this.$searchInputs.forEach((input) => {
```
aleksandraworhacz: What will happen if there are multiple search fields? Wouldn't this code cause focus to jump among them?

aleksandraworhacz (approved):
Looks good :)

PR #127, opened 2024-03-19
Xcorner 2 mega menu qa feedback

mgbelegu:
- added show more functionality on mobile submenu items
- fixed social icons styling


assets/js/theme/custom/mega-menu.js
```diff
@@ -0,0 +1,58 @@
...
+export default function megaMenu() {
+    const MOBILE_WIDTH = 768
+    const megaMenuElement = document.querySelector('.js-navigation-megamenu')
+    const searchButton = document.querySelector('.js-mega-menu-search')
+    const searchInputs = document.querySelectorAll('.js-search-quick')
+    const showMore = document.querySelectorAll('.js-show-more')
+    let hasAttachedEvent = false
+
+    const openDesktopMenu = () => {
+        megaMenuElement.previousElementSibling.classList.toggle('is-open')
```
aleksandraworhacz: This could be accidentally broken by basic html edit. Please add a dedicated js classname for the element instead of using `previousElementSibling`

assets/js/theme/custom/mega-menu.js
```diff
@@ -0,0 +1,58 @@
+/**
+ * Open mega menu on search button click
+ */
+export default function megaMenu() {
+    const MOBILE_WIDTH = 768
+    const megaMenuElement = document.querySelector('.js-navigation-megamenu')
+    const searchButton = document.querySelector('.js-mega-menu-search')
+    const searchInputs = document.querySelectorAll('.js-search-quick')
+    const showMore = document.querySelectorAll('.js-show-more')
```
aleksandraworhacz: Since there could be multiple elements selected, maybe rename to `showMoreTriggers` ?

assets/js/theme/custom/mega-menu.js
```diff
@@ -0,0 +1,58 @@
...
+    let hasAttachedEvent = false
+
+    const openDesktopMenu = () => {
+        megaMenuElement.previousElementSibling.classList.toggle('is-open')
+        megaMenuElement.classList.toggle('is-open')
+        searchInputs?.forEach((input) => {
```
aleksandraworhacz: Iterating through search Inputs and changing focus on each iteration doesn't seem right. Or is it intentional?

assets/js/theme/custom/mega-menu.js
```diff
@@ -0,0 +1,58 @@
...
+        }
+    }
+
+    updateEventListeners()
+
+    window.addEventListener('resize', updateEventListeners)
+
+    showMore.forEach((button) => {
+        button.addEventListener('click', () => {
+            const buttonLabels = button.querySelectorAll('span')
```
aleksandraworhacz: Please use ".js-..." classname instead of selecting all `span`s
Also, why do you need multiple labels to select? Shouldn't there be one label?

assets/js/theme/custom/mega-menu.js
```diff
@@ -0,0 +1,58 @@
...
+            const svg = button.querySelector('svg')
```
aleksandraworhacz: Do you need this? Couldn't you rotate the icon only with styles? Something like:

```
.icon-selector {
  .some-parent-selector.is-open & {
      rotate code here
  }
}
```

assets/js/theme/custom/mega-menu.js
```diff
@@ -0,0 +1,58 @@
...
+            const liElements = []
+            let prevElement = button.parentElement.previousElementSibling
+
+            while (prevElement) {
```
aleksandraworhacz: Why couldn't you use `js` classname for elements that are hidden instead of relying on `previousElementSibling`? If the structure slightly changes it will crash.

assets/js/theme/custom/mega-menu.js
```diff
@@ -0,0 +1,58 @@
...
+                if (!prevElement.classList.contains('c-navigation-list__item--is-shown')) {
```
aleksandraworhacz: Please don't rely on "regular" classname in js as they may change, use `js` ones instead

assets/js/theme/custom/mega-menu.js
```diff
@@ -0,0 +1,58 @@
...
+                    prevElement.classList.toggle('u-hidden@medium-down')
```
aleksandraworhacz: Don't toggle `u-hidden` classname,  instead toggle `is-hidden` and handle its styling in stylesheet - much easier to maintain in the future

assets/js/theme/custom/mega-menu.js
```diff
@@ -0,0 +1,58 @@
...
+                }
+                liElements.push(prevElement)
+                prevElement = prevElement.previousElementSibling
+            }
+
+            if (liElements.length > 0) {
+                buttonLabels.forEach((label) => {
+                    label.classList.toggle('u-hidden')
```
aleksandraworhacz: You don't need to toggle visibility class on each item inside the button (labels, SVG), you could toggle `is-open` on the button, and style other elements in scss based on `is-open`

e.g.

```
.label-class {
  .button.is-open & { ...code for shown label} 
}
```

assets/js/theme/global/mobile-menu-toggle.js
```diff
@@ -35,6 +35,7 @@ export class MobileMenuToggle {
         this.$toggle = $toggle
         this.$searchToggle = q$('.js-mobile-mega-menu-search')
         this.$searchInput = q$('.js-mobile-search')
+        this.$bodyOverlay = q$('.c-body-overlay')
```
aleksandraworhacz: Use `js` classname

templates/components/common/navigation-menu.html
```diff
@@ -22,7 +22,7 @@
     <div class="c-navigation__megamenu is-mobile">
         <div class="o-container">
             <div class="js-quick-search">
-                {{> components/common/quick-search name='nav-menu-quick-search' mobile="yes"}}
+                {{> components/common/quick-search name='nav-menu-quick-search' mobile='yes'}}
```
aleksandraworhacz: Consider using boolean param instead of string "yes"

templates/components/common/navigation-menu.html
```diff
@@ -49,6 +49,15 @@
                                             </a>
                                         </li>
                                     {{/each}}
+                                    <li>
+                                        <button class="js-show-more c-navigation-list__show-more">
+                                            <span>{{lang 'common.show_more'}}</span>
+                                            <span class="u-hidden">{{lang 'common.show_less'}}</span>
+                                            <svg class="c-navigation-list__arrow">
```
aleksandraworhacz: Maybe use `components/icon` ?

assets/scss/components/_navigation.scss
```diff
@@ -122,6 +127,39 @@
...
+}
+
+.c-navigation-list__arrow {
+    fill: $color-white;
+    transform: rotate(0deg);
+    transition: all 0.2s ease-in-out;
+    max-width: rem-calc(24px);
+    max-height: rem-calc(24px);
+
+    &.rotate {
```
aleksandraworhacz: Try using adjectives instead of verbs :)

assets/js/theme/custom/mega-menu.js
```diff
@@ -0,0 +1,58 @@
+/**
+ * Open mega menu on search button click
+ */
+export default function megaMenu() {
+    const MOBILE_WIDTH = 768
```
aleksandraworhacz: Is this value used anywhere else? Maybe it would be good to keep javascript breakpoint values in one place for the sake of consistency?

aleksandraworhacz (commented):
Comments added

assets/js/theme/custom/mega-menu.js
mgbelegu: There are two labels inside the button. We toggle both span elements with the classname "u-hidden" where one gets hidden and the other appears. I didn't want to change the value by hardcoding into HTML, so that we can use {{lang}} translated strings:

<span>Show more</span>
<span class="u-hidden">Show less</span>

assets/js/theme/custom/mega-menu.js
mgbelegu: @aleksandraworhacz What is the difference?

mgbelegu:
@aleksandraworhacz Hi, can you please review again? I've also asked a question in one of the comments. Thank you

assets/js/theme/custom/mega-menu.js
```diff
@@ -0,0 +1,55 @@
+/**
+ * Open mega menu on search button click
+ */
+
+import { constants } from './constants'
+
+export default function megaMenu() {
+    const MOBILE_WIDTH = constants.MOBILE_WIDTH
```
aleksandraworhacz: Use destructing 

```
const {MOBILE_WIDTH} = contstants;
```

assets/js/theme/custom/mega-menu.js
```diff
@@ -0,0 +1,55 @@
...
+    window.addEventListener('resize', updateEventListeners)
+
+    showMoreTriggers.forEach((item) => {
+        const button = item.querySelector('.js-show-more')
+        button.addEventListener('click', () => {
+            button.classList.toggle('is-open')
+
+            const listItems = item
+                .closest('.js-navigation-list-container')
+                .querySelectorAll('.js-navigation-list-item:not(.js-navigation-list-item-shown)')
```
aleksandraworhacz: This seems overcomplicated. You could use `data` attribute and get rid of climbing up the DOM. Since you display `showMore` button and a list of links inside a loop, you could set `data-show-more="CATEGORY_ID"` and `data-more="CATEGORY_ID"` and then on click,  just query for links with `data-more` value the same as button's `data-show-more`

Also, I recommend to avoid toggling utility classes like `u-hidden@medium-down`, in general it's better to mark element as hidden with `is-hidden` and then handle its visibility directly in styles.

There's a higher chance that someone updates the utility's styles and unintentionally break the styling (not being aware that this class is being toggle somewhere by js) than accidentally change styles in component.

assets/js/theme/custom/mega-menu.js
```diff
@@ -0,0 +1,55 @@
...
+        }
+    }
+
+    updateEventListeners()
+
+    window.addEventListener('resize', updateEventListeners)
+
+    showMoreTriggers.forEach((item) => {
+        const button = item.querySelector('.js-show-more')
+        button.addEventListener('click', () => {
```
aleksandraworhacz: It's worth to check if button exists before adding event listener

aleksandraworhacz (commented):
Almost there :)

assets/js/theme/custom/mega-menu.js
```diff
@@ -0,0 +1,57 @@
...
+                button.classList.toggle('is-open')
+
+                const buttonLabels = button.querySelectorAll('.js-show-label')
+                if (listItems.length > 0) {
+                    listItems.forEach((listItem) => {
+                        listItem.classList.toggle('is-hidden')
+                    })
+
+                    buttonLabels.forEach((label) => {
+                        label.classList.toggle('u-hidden')
```
aleksandraworhacz: use `is-hidden`  instead

assets/js/theme/custom/mega-menu.js
```diff
@@ -0,0 +1,57 @@
...
+                    })
+                }
+            })
+
+            button.setAttribute('data-event-added', 'true')
```
aleksandraworhacz: What's the purpose of `data-event-added` ?

mgbelegu: So that the event listener is not added twice to the button

aleksandraworhacz: But is it even possible that the event will be added twice? You call `megaMenu()` once, so you don't really need to worry about it in this case?

Also, consider wrapping `showMoreTriggers.forEach ....` code with `function`, it will look more clear :)

mgbelegu: Well not sure why when I hard refresh or when I resize it seems like the event listener is added twice and it doesn't let me "hide" the menu items, because it is toggling 'is-hidden' twice. Meaning it is removing and then adding the classname again. It means the on click event is twice in the button element. You can give it a try yourself without that 'date-event-added' code

aleksandraworhacz: That's something we may need to investigate, as it could cause troubles in the future. But, even tho, why can't you just create a variable outside `megaMenu` called `isShowMoreEventAttached = false` and then assign it to true after finish iterating through forEach loop? Seems like much cleaner approach, than setting `data-event-added` attribute

assets/js/theme/custom/mega-menu.js
```diff
@@ -0,0 +1,61 @@
...
+
+import { constants } from './constants'
+
+export default function megaMenu() {
+    const { MOBILE_WIDTH } = constants
+    const dropdown = document.querySelector('.js-dropdown')
+    const searchButton = document.querySelector('.js-mega-menu-search')
+    const searchInput = document.querySelector('.js-search-quick')
+    const showMoreTriggers = document.querySelectorAll('.js-show-more-item')
+    let hasAttachedEvent = false
```
aleksandraworhacz: Since the `megaMenu` is called twice, you can't rely on this variable as it will be set to `false` each time the megaMenu is being called. I suggest to move it outside the function, and apply similar solution for the `showMoreMenuItems` - using `data-event-added` isn't a best solution as you may need to attach multiple events, and in such case the `data-event-added` won't have any meaning.

mgbelegu: I've updated it. Please check again.  By using button.dataset.hasShowMoreEvent = 'true', I can track each button's event listener attachment directly on the element itself. This ensures that we don't unintentionally block event listeners from being added to other "Show More" buttons while still preventing duplicates on the same button.

aleksandraworhacz: Is it possible to not attach an event to all buttons if you are calling `addEventListener` inside a loop?

assets/js/theme/custom/mega-menu.js
```diff
@@ -0,0 +1,63 @@
...
+                searchButton.addEventListener('click', openDesktopMenu)
+                hasEvent = true
+            }
+        }
+    }
+
+    const showMoreMenuItems = () => {
+        showMoreTriggers.forEach((item) => {
+            const button = item.querySelector('.js-show-more')
+            if (button && !isShowMoreEventAttached) {
```
aleksandraworhacz: if the `isShowMoreEventAttached` is `true` I think we could skip iterating through buttons at all?

assets/js/theme/custom/mega-menu.js
```diff
@@ -0,0 +1,63 @@
+/**
+ * Open mega menu on search button click
+ */
+
+import { constants } from './constants'
+
+let isShowMoreEventAttached = false
+
+export default function megaMenu(hasAttachedEvent) {
+    let hasEvent = hasAttachedEvent
```
aleksandraworhacz: Naming is a little confusing here and not descriptive and concrete enough

aleksandraworhacz (approved):
Approved 👍

PR #130, opened 2024-03-22
Xcorner 7 truncate reviews

mgbelegu:


aleksandraworhacz (approved):
Looks good, however I'd consider creating a mixin as this functionality could be helpful in other places :) 

Also, it's rather a design issue than code one, but if we truncate a review, how users could able to read it? 🤔

PR #133, opened 2024-03-26
QA feedback

mgbelegu:


assets/scss/components/_form.scss
```diff
@@ -198,6 +198,14 @@ $input-transition: all 0.2s ease-in-out;
         width: 100%;
     }
 
+    &--full-mobile {
```
aleksandraworhacz: to keep it consistent, maybe rename to `&--full\@mobile` ?

aleksandraworhacz (approved):
One tiny comment added, but besides that approved :)

PR #137, opened 2024-03-27
Xcorner 13 qa feedback

mgbelegu:
- fixed scroll to top when search page loads
- fixed search url when searching in other pages other than home

### DIFF
--- a/assets/js/theme/global/quick-search.js
+++ b/assets/js/theme/global/quick-search.js
@@ -96,9 +96,9 @@ export default function doQuickSearch() {
                 return
             }
 
-            const protocol = window.location.protocol
-            const host = window.location.host
-            window.location.href = `${protocol}//${host}/search.php?search_query=${encodeURIComponent(searchQuery)}`
+            const origin = window.location.origin
+
+            window.location.href = `${origin}/search.php?search_query=${encodeURIComponent(searchQuery)}`
         })
     })
 }
--- a/assets/js/theme/global/quick-search.js
+++ b/assets/js/theme/global/quick-search.js
@@ -96,7 +96,9 @@ export default function doQuickSearch() {
                 return
             }
 
-            window.location.href = `search.php?search_query=${encodeURIComponent(searchQuery)}`
+            const protocol = window.location.protocol
+            const host = window.location.host
+            window.location.href = `${protocol}//${host}/search.php?search_query=${encodeURIComponent(searchQuery)}`
         })
     })
 }
--- a/assets/js/theme/search.js
+++ b/assets/js/theme/search.js
@@ -223,8 +223,6 @@ export default class Search extends CatalogPage {
 
         requestAnimationFrame(() => {
             $searchResultsMessage.focus()
-            // document.querySelector('html').scrollTop = 0
-
             window.scrollTo(0, 0)
         })
     }
--- a/assets/js/theme/search.js
+++ b/assets/js/theme/search.js
@@ -221,11 +221,10 @@ export default class Search extends CatalogPage {
 
         q$('body').insertAdjacentElement('beforeend', $searchResultsMessage)
 
-        setTimeout(() => {
+        requestAnimationFrame(() => {
             $searchResultsMessage.focus()
-            q$('html').scrollTop = 0
-            q$('body').scrollTop = 0
-        }, 100)
+            document.querySelector('html').scrollTop = 0
+        })
     }
 
     loadTreeNodes(node, cb) {
--- a/assets/js/theme/search.js
+++ b/assets/js/theme/search.js
@@ -223,7 +223,9 @@ export default class Search extends CatalogPage {
 
         requestAnimationFrame(() => {
             $searchResultsMessage.focus()
-            document.querySelector('html').scrollTop = 0
+            // document.querySelector('html').scrollTop = 0
+
+            window.scrollTo(0, 0)
         })
     }
 
--- a/assets/js/theme/search.js
+++ b/assets/js/theme/search.js
@@ -221,7 +221,11 @@ export default class Search extends CatalogPage {
 
         q$('body').insertAdjacentElement('beforeend', $searchResultsMessage)
 
-        setTimeout(() => $searchResultsMessage.focus(), 100)
+        setTimeout(() => {
+            $searchResultsMessage.focus()
+            q$('html').scrollTop = 0
+            q$('body').scrollTop = 0
+        }, 100)
     }
 
     loadTreeNodes(node, cb) {
--- a/assets/scss/components/_announcement-banner.scss
+++ b/assets/scss/components/_announcement-banner.scss
@@ -1,0 +1,34 @@
+.c-announcement-banner {
+    display: flex;
+    justify-content: center;
+    align-items: center;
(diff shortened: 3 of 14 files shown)


assets/js/theme/global/quick-search.js
```diff
@@ -96,7 +96,9 @@ export default function doQuickSearch() {
                 return
             }
 
-            window.location.href = `search.php?search_query=${encodeURIComponent(searchQuery)}`
+            const protocol = window.location.protocol
+            const host = window.location.host
```
aleksandraworhacz: Why not using `window.location.origin` instead? 🤔

assets/js/theme/search.js
```diff
@@ -221,7 +221,11 @@ export default class Search extends CatalogPage {
 
         q$('body').insertAdjacentElement('beforeend', $searchResultsMessage)
 
-        setTimeout(() => $searchResultsMessage.focus(), 100)
+        setTimeout(() => {
+            $searchResultsMessage.focus()
+            q$('html').scrollTop = 0
+            q$('body').scrollTop = 0
```
aleksandraworhacz: Why do you need to scroll both `html` and `body` ?  If you need to execute some code with a slightly delay, try use `requestAnimationFrame` first instead of using setTimeout with magic delay value - I know that you just added `scrollTop` tho

Also, I encourage to use vanilla js code instead of jquery :)

templates/components/common/footer.html
```diff
@@ -1,6 +1,6 @@
 {{#if banners.bottom}}
     {{#if settings.data_tag_enabled}}
-        <div class="c-header__announcement-banner" data-banner-location="bottom">
+<div class="c-header__announcement-banner" data-banner-location="bottom" xmlns="http://www.w3.org/1999/html">
```
aleksandraworhacz: Should `c-header__announcement-banner`be used inside `footer` component? 🤯

templates/pages/search.html
```diff
@@ -18,56 +18,56 @@
...
-            <span class="c-search__hide-filters-label js-hide-filters-label">{{lang 'search.faceted.hide-filters'}}</span>
-            <span class="c-search__hide-filters-label js-hide-filters-label is-open">{{lang 'search.faceted.show-filters'}}</span>
-            <svg width="15" height="8" viewBox="0 0 15 8" fill="none" xmlns="http://www.w3.org/2000/svg">
-                <path d="M13.5 7L7.5 1L1.5 7" stroke="#5A64C1" stroke-width="1.5" stroke-linecap="round" stroke-linejoin="round"/>
-            </svg>
-        </button>
+    <button class="c-button c-button--secondary c-search__hide-filters js-search-hide-filters">
+        <span class="c-search__hide-filters-label js-hide-filters-label">{{lang 'search.faceted.hide-filters'}}</span>
+        <span class="c-search__hide-filters-label js-hide-filters-label is-open">{{lang 'search.faceted.show-filters'}}</span>
+        <svg width="15" height="8" viewBox="0 0 15 8" fill="none" xmlns="http://www.w3.org/2000/svg">
```
aleksandraworhacz: SVG sprite + icon compoent maybe? :D

aleksandraworhacz (commented):
Comments added :)

assets/js/theme/search.js
mgbelegu: I love vanilla JS better too, just wanted to keep it consistent. But i'll change it no worries

mgbelegu:

assets/js/theme/search.js
```diff
@@ -221,7 +221,10 @@ export default class Search extends CatalogPage {
 
         q$('body').insertAdjacentElement('beforeend', $searchResultsMessage)
 
-        setTimeout(() => $searchResultsMessage.focus(), 100)
+        requestAnimationFrame(() => {
+            $searchResultsMessage.focus()
+            document.querySelector('html').scrollTop = 0
```
aleksandraworhacz: You could scroll directly the window like `window.scrollTo(0,0);` :)

assets/scss/components/_announcement-banner.scss
```diff
@@ -0,0 +1,34 @@
+.c-announcement-banner {
+    display: flex;
+    justify-content: center;
+    align-items: center;
+    background-color: $color-black;
+    padding: 1rem 1.5rem;
+    position: relative;
+    z-index: $z-index-mid;
+
+    p {
```
aleksandraworhacz, line 10: is the announcement bar as Wysiwyg container? If not, could we have a dedicated class names for the nested elements?

assets/scss/components/_search.scss
```diff
@@ -86,3 +86,13 @@
         display: block;
     }
 }
+
+.c-search__hide-filters-icon {
```
aleksandraworhacz: in `c-search__hide-filters-icon` the part `hide-filters-icon` isn't quite right considering the BEM methodology. It would be better to name it `c-search__icon--hidden-filters` or `c-search__icon--no-filters`

assets/js/theme/search.js
```diff
@@ -221,7 +221,12 @@ export default class Search extends CatalogPage {
 
         q$('body').insertAdjacentElement('beforeend', $searchResultsMessage)
 
-        setTimeout(() => $searchResultsMessage.focus(), 100)
+        requestAnimationFrame(() => {
+            $searchResultsMessage.focus()
+            // document.querySelector('html').scrollTop = 0
```
aleksandraworhacz: Remove commented line :)

assets/scss/components/_search.scss
```diff
@@ -86,3 +86,13 @@
         display: block;
     }
 }
+
+.c-search__icon--hide-filters {
+    fill: transparent;
+
+    .c-search__hide-filters:hover,
```
aleksandraworhacz: `__hide-filters` doesn't seem to be a proper name for the element in BEM, so maybe use it rather as modifier?
`.c-search--hidden-filters` ?

aleksandraworhacz (approved):
Added two mini comments, but overall looks good :) approved 👍

assets/scss/components/_button.scss
```diff
@@ -59,6 +59,34 @@
     padding: 0.75rem 1rem;
     margin: 0;
     cursor: pointer;
+
+    &--with-icon {
```
aleksandraworhacz, line 63: just `&--icon` would work as well, no need to overcomplicating the naming :)

assets/scss/components/_button.scss
```diff
@@ -59,6 +59,34 @@
...
+                stroke: white;
+                transition: all 0.2s ease-in-out;
+            }
+        }
+
+        svg {
+            transform: rotate(180deg);
+            transition: all 0.2s ease-in-out;
+
+            &.is-open {
```
aleksandraworhacz: it would make more sense to toggle `is-open` directly on button, instead of the SVG. Then you could do
```
&--icon {
  &.is-open svg { .... }
}
```

assets/scss/components/_button.scss
mgbelegu: there is another c-button--icon classname specified. i think we can keep c-button--with-icon

PR #141, opened 2024-04-05
XCORNER-39 checkout styling

mgbelegu:
- imported styles on the optimized checkout
- extended styling to default checkout classes to match the design


assets/scss/optimized-checkout.scss
```diff
@@ -1 +1,111 @@
...
+@import 'vendor/vendor';
+@import 'objects/objects';
+@import 'components/components';
+@import 'utilities/utilities';
+
+@mixin form-input() {
+    box-shadow: none;
+}
+
+.layout,
```
aleksandraworhacz, line 11: Please add classname prefix if possible e.g. `.c-...`

assets/scss/optimized-checkout.scss
```diff
@@ -1 +1,111 @@
...
+@import 'settings/settings';
+@import 'tools/tools';
+@import 'generic/generic';
+@import 'elements/elements';
+@import 'vendor/vendor';
+@import 'objects/objects';
+@import 'components/components';
+@import 'utilities/utilities';
+
+@mixin form-input() {
```
aleksandraworhacz: Is it used in multiple places or just at line 40? If it's used in one place, you don't really need a mixin

aleksandraworhacz (commented):
Added few comments

assets/scss/optimized-checkout.scss
mgbelegu: @aleksandraworhacz The checkout HTML content is rendered by BigCommerce and we have no control over it. We can't change the classnames or HTML.

(unless we use the BigCommerce Checkout SDK, but thats another story 😄  )

assets/scss/optimized-checkout.scss
mgbelegu: Done

mgbelegu:
@aleksandraworhacz Can you please check again?

assets/scss/optimized-checkout.scss
aleksandraworhacz: I thought so, but it was worth to ask :)

aleksandraworhacz (approved):

PR #142, opened 2024-04-05
[XCORNER-26]: Account pages

aleksandraworhacz:
* Make c-alert and wishlist's buttons full width on mobile
* Prevent displaying messages list when empty


mgbelegu (approved):
Looks good :)

PR #145, opened 2024-04-28
XCORNER-50 change cart layout

mgbelegu:


assets/scss/components/_cart.scss
```diff
@@ -1,7 +1,22 @@
...
+
 .c-cart__totals {
     display: flex;
     flex-direction: column;
     align-items: flex-end;
+    position: sticky;
+    position: -webkit-sticky;
+    top: 20px;
+    height: min-content;
+    z-index: 10;
```
aleksandraworhacz: Consider putting the `z-index` into variable. It would make it easier to maintain all z-indexes in the future :)

aleksandraworhacz (approved):
One small suggestion added, but approved :)

assets/scss/components/_cart.scss
mgbelegu: I actually removed it, we don't need it anymore :)

PR #146, opened 2024-04-28
XCORNER-48 reload the page when removing all products from compare page

mgbelegu:


assets/js/theme/global/compare-products.js
```diff
@@ -32,6 +32,10 @@ function decrementCounter(itemID) {
         const url = new URL(pathname)
         window.history.replaceState(null, '', url)
         q$(`[${selectors.dataCompareCard}='${itemID}']`).remove()
+
+        if (compareProductsIDs.length === 0) {
+            window.location.reload()
```
aleksandraworhacz: Hmm, why do you need to reload the page? 🤔
It will cause the loading of all resources again, which is not great from the performance POV.

mgbelegu: @aleksandraworhacz This is to restore the page to the initial view when it doesn't have any products. Any suggestions in mind?

aleksandraworhacz: How about removing `.c-product-compare` update heading's content or toggle some `visibility class` on the container, to display either "no results" heading or "product-compare" container? 

You wouldn't need to reload the page in such case 🤔

mgbelegu:

assets/js/theme/global/compare-products.js
```diff
@@ -32,6 +34,12 @@ function decrementCounter(itemID) {
         const url = new URL(pathname)
         window.history.replaceState(null, '', url)
         q$(`[${selectors.dataCompareCard}='${itemID}']`).remove()
+
+        if (compareProductsIDs.length === 0) {
```
aleksandraworhacz: This code isn't strictly related to the `decrementCounter` function. Let's move it into a dedicated one called `handleNoResults` (or something like that). Also, handling visibility by CSS seems easier - as you wouldn't need to select multiple elements via js, and update their styles individually. 

So, how about adding a class "no-results" to the container, and based on this, display or hide compare products / remove all button and a proper content for h1? 🤔

mgbelegu:
@aleksandraworhacz Changed! Can you please review again?

assets/scss/components/_product-compare.scss
```diff
@@ -1,5 +1,25 @@
 $details-border: 1px solid $color-neutrals-75;
 
+.c-products-header {
+    display: block;
+
+    .no-results & {
+        display: none;
+    }
+
+    &--empty {
```
aleksandraworhacz, line 10: It's more a suggestion, but instead hading two h1 tags, you cold have two spans inside h1 tag and toggle spans visibility instead of having more complicated code like this :D

aleksandraworhacz (approved):
Looks good :)

PR #147, opened 2024-04-28
XCORNER-41 make nav menu fixed

mgbelegu:


assets/js/theme/global/menu.js
```diff
@@ -75,6 +82,52 @@ class Menu {
...
+    updateBodyMargin() {
+        if (this.$header.classList.contains('c-header__fixed')) {
+            const headerHeight = this.$header.offsetHeight
+            this.$body.style.marginTop = `${headerHeight}px`
+        } else {
+            this.$body.style.marginTop = '0px'
+        }
+    }
+
+    onScroll() {
```
aleksandraworhacz, line 95: wooah! you overcomplicated it 😅 

1. Let's use `position: sticky` instead of having so much fun with toggling `position: fixed` and adjust body's spacing. Simple `sticky` does it for you.
2. The Y movement of a header could be done a bit simpler:
  a. On init and on "resize" update CSS variable `--navigation-height` that will contain a height of `c-nagivation` element.
  b. then you could easily move the entire header like:

```
[scss file]
.c-header.pushed {
   transform: translateY(calc(-100% + var(--navigation-height, 0px)));
}
--------------
[js]

onScroll() {
  const operation = currentScrollTop > this.lastScrollTop ? 'add' : 'remove';
  this.$header.classList[operation]('pushed');
  this.lastScrollTop = currentScrollTop
}
```

aleksandraworhacz (commented):
Comment added

mgbelegu:
@aleksandraworhacz Hi, can you please review it again? I've changed it :)

assets/scss/components/_menu.scss
```diff
@@ -34,6 +34,18 @@
     }
 }
 
+.c-menu__fixed {
```
aleksandraworhacz: I don't think you need this class anymore :)

assets/scss/components/_navigation.scss
```diff
@@ -242,6 +247,7 @@
     flex-direction: column;
     justify-content: center;
     align-items: center;
+    margin-top: 216px;
```
aleksandraworhacz: `rem-calc()` maybe? :)

aleksandraworhacz (approved):
Few minor comments added but overall approved :) 
Looks much much better now 🔥

PR #148, opened 2024-04-29
XCORNER-46 add wishlist popup modal

mgbelegu:


assets/js/theme/custom/wishlist-modal.js
```diff
@@ -0,0 +1,38 @@
+import modalFactory from '../global/modal'
+
+export default function wishlistModal() {
+    const selectors = {
+        wishlist: '.js-wishlist-add',
+    }
+
+    const text = {
```
aleksandraworhacz: How it would be handled for other languages? 🤔

assets/js/theme/custom/wishlist-modal.js
```diff
@@ -0,0 +1,38 @@
...
+        message: 'Please login or register to add this item to your wishlist.',
+        login: 'Login',
+        createAccount: 'Create Account',
+    }
+
+    const wishlistButtons = document.querySelectorAll(selectors.wishlist)
+    wishlistButtons?.forEach((button) => {
+        const isCustomer = button.getAttribute('customer')
+        if (isCustomer !== 'null') return
+        button?.addEventListener('click', (ev) => {
```
aleksandraworhacz: I think at this point the button need to exists

assets/js/theme/custom/wishlist-modal.js
```diff
@@ -0,0 +1,38 @@
...
+            ev.preventDefault()
+            const modal = modalFactory('.js-preview-modal')
+            if (modal) {
+                modal.open()
+
+                const contentHTML = `
```
aleksandraworhacz: I'd suggest having two "add to wishlist buttons" (if you need to handle wishlists for guests and logged-in users) and displaying one of those buttons—but directly in HTML — based on the customer variable. It would be easier to handle multilingualism this way.

aleksandraworhacz (changes requested):

mgbelegu:
@aleksandraworhacz Please review again :)

templates/components/common/wishlist-add.html
```diff
@@ -1,7 +1,29 @@
...
-    </button>
-</form>
+{{#if customer '!==' null}}
+    <form 
+        action="{{add_to_wishlist_url}}" 
+        class="js-wishlist-add c-wishlist" 
+        method="post"
+    >
+        <button class="c-wishlist__button" type="submit">
+            <svg xmlns="http://www.w3.org/2000/svg" width="24" height="24" viewBox="0 0 24 24" fill="none">
```
aleksandraworhacz: Why not  to use`components/icon` ? 🤔

templates/components/common/wishlist-add.html
```diff
@@ -1,7 +1,29 @@
...
+        action="{{add_to_wishlist_url}}" 
+        class="js-guest-wishlist-add c-wishlist" 
+        method="post"
+        data-modal-title="{{lang 'account.wishlists.login_to_add'}}"
+        data-modal-message="{{lang 'account.wishlists.login_to_add_message'}}"
+        data-modal-login="{{lang 'login.heading'}}"
+        data-modal-create-account="{{lang 'login.new_customer.create_account'}}"
+    >
+        <button class="c-wishlist__button" type="submit">
+            <svg xmlns="http://www.w3.org/2000/svg" width="24" height="24" viewBox="0 0 24 24" fill="none">
```
aleksandraworhacz: same here - `components/icon` ? :)

assets/js/theme/custom/wishlist-modal.js
```diff
@@ -0,0 +1,36 @@
...
+
+        if (button) {
+            button.addEventListener('click', (ev) => {
+                ev.preventDefault()
+                const modal = modalFactory('.js-preview-modal')
+                if (modal) {
+                    modal.open()
+
+                    const contentHTML = `
+                    <h1 class="c-modal__title">${title}</h1>
```
aleksandraworhacz, line 23: Ok, it does solve the multi-language issue (I give you that 😅 ), but do we really need to inject it via js? Why can't we have a modal with the content inside html and just toggle its visibility via js? 🤔

aleksandraworhacz (commented):
Comments added :)

templates/components/common/wishlist-add.html
mgbelegu: This was implemented before the components/icon change and just forgot to replace it. Will do so now

assets/js/theme/custom/wishlist-modal.js
mgbelegu: I didn't want to add another modal element in the HTML and the way I'm doing it now is just using what we already have

aleksandraworhacz: `I didn't want to add another modal element in the HTML` why not? 🤔

assets/scss/components/_wishlist-list.scss
```diff
@@ -31,3 +31,12 @@
     gap: 12px;
     align-items: center;
 }
+
+.c-wishlist__icon {
```
aleksandraworhacz: this should be called `c-whishlist-list__icon` to be consistent with the file's name and the rest of classnames :)

assets/js/theme/custom/wishlist-modal.js
mgbelegu: @aleksandraworhacz Because that is how the modal currently works and I wanted to keep it consistent. What if in the future we keep adding modals? We will just have hidden HTML content on the page at all times and we will trigger it's visibility via JS. Having one modal HTML div where we can add content to it will reduce DOM complexity and page load time. Also, having multiple html modals seems redundant to me

aleksandraworhacz: Well, I won't argue on having it consistent cause you have a point on that. But having more complex DOM structure isn't an issue, actually it's simpler than modifying the structure with js, so in terms of complexity I'd say the less js code the better and faster.  

>  Also, having multiple html modals seems redundant to me
How is it different than having multiple modal files but with js code? I think it's easier to maintain the code if you have dedicated HTML files for HTML components than searching for HTML code inside js file only in specific cases. 

In the end you decide what approach work best, just leaving here my humble opinion :)

assets/scss/components/_wishlist-list.scss
```diff
@@ -31,3 +31,12 @@
     gap: 12px;
     align-items: center;
 }
+
+.c-wishlist-list__icon {
+    width: rem-calc(24px);
+    height: rem-calc(24px);
+
+    use {
+        fill: transparent;
```
aleksandraworhacz: You can set `fill: transparent` directly in `c-wishlist-list__icon`, no need to nest `use` here

mgbelegu:
@aleksandraworhacz  I'd rather keep it as it is with injecting JS. I've already spent too much time on these tasks and since time is very valuable here I would rather skip the amendments for now

aleksandraworhacz (approved):
Approved (however injecting HTML could be done differently 😉 )

PR #149, opened 2024-04-29
XCORNER-47 compare popup modal

mgbelegu:


assets/js/theme/custom/compare-modal.js
```diff
@@ -0,0 +1,39 @@
...
+
+    const compareInputs = document.querySelectorAll(selectors.compareInputs)
+    compareInputs?.forEach((input) => {
+        input?.addEventListener('click', () => {
+            if (!input.checked) return
+            const modal = modalFactory('.js-preview-modal')
+            if (modal) {
+                modal.open()
+
+                const contentHTML = `
```
aleksandraworhacz, line 19: Make it more "multilanguage friendly" :)

assets/js/theme/custom/wishlist-modal.js
```diff
@@ -0,0 +1,38 @@
+import modalFactory from '../global/modal'
+
+export default function wishlistModal() {
+    const selectors = {
+        wishlist: '.js-wishlist-add',
+    }
+
+    const text = {
```
aleksandraworhacz: Same here

aleksandraworhacz (changes requested):

templates/components/common/wishlist-add.html
```diff
@@ -1,4 +1,9 @@
-<form action="{{add_to_wishlist_url}}" class="js-wishlist-add c-wishlist" method="post">
+<form 
+    action="{{add_to_wishlist_url}}" 
+    class="js-wishlist-add c-wishlist" 
+    method="post" 
+    customer="{{#if customer '!==' null}}{{customer}}{{else}}null{{/if}}"
+>
     <button class="c-wishlist__button" type="submit">
         <svg xmlns="http://www.w3.org/2000/svg" width="24" height="24" viewBox="0 0 24 24" fill="none">
```
aleksandraworhacz: `components/icon` ? :)

assets/js/theme/custom/compare-modal.js
aleksandraworhacz: I'm not a fan of injecting the content via HTML. But if there's absolutely no way to handle this in a more traditional way (content in HTML, toggling visibility in JS), then it's fine.

assets/js/theme/global/menu.js
```diff
@@ -75,6 +82,27 @@ class Menu {
     onDocumentClick() {
         this.collapseAll()
     }
+
+    updateBodyMargin() {
+        if (this.$header.classList.contains('c-header__fixed')) {
+            const headerHeight = this.$header.offsetHeight
+            this.$body.style.marginTop = `${headerHeight}px`
+        } else {
+            this.$body.style.marginTop = '0px'
```
aleksandraworhacz: you can set `this.$body.style.marginTop = ''` to reset :)

assets/scss/components/_cart.scss
```diff
@@ -1,7 +1,25 @@
...
+    }
+}
+
 .c-cart__totals {
     display: flex;
     flex-direction: column;
     align-items: flex-end;
+    position: sticky;
+    position: -webkit-sticky;
+    top: 20px;
```
aleksandraworhacz: Should it be fixed or rem related?

assets/scss/components/_header.scss
```diff
@@ -6,6 +6,16 @@
...
+    top: 0;
+    transition: transform 0.3s ease-in-out;
+    z-index: $z-index-high;
+}
+
+.c-header.pushed {
+    transform: translateY(-100%);
+
+    @include bp(medium) {
+        transform: translateY(calc(-100% + 54px));
```
aleksandraworhacz: I'd suggest to avoid using "magic numbers" and put it into a variable with a descriptive name

assets/scss/components/_navigation.scss
```diff
@@ -242,6 +248,14 @@
     flex-direction: column;
     justify-content: center;
     align-items: center;
+
+    @include bp(medium) {
+        margin-top: rem-calc(432px);
```
aleksandraworhacz: Does this value have a meaning (e.g. is it related to another element)? If it does, maybe consider moving it to a variable?

templates/pages/cart.html
```diff
@@ -12,68 +12,70 @@
...
-                                {{lang 'cart.preview.checkout_multiple'}}
-                            </a>
+                        {{#if cart.show_primary_checkout_button}}
+                            <div class="c-cart__checkout-button-wrapper">
+                                <a
+                                    class="c-button c-button--primary c-totals__checkout-button c-button--full "
+                                    href="{{urls.checkout.single_address}}"
+                                    title="{{lang 'cart.checkout.title'}}"
+                                >
+                                    <svg width="24" height="24" viewBox="0 0 24 24" fill="none" xmlns="http://www.w3.org/2000/svg">
```
aleksandraworhacz: Could it be replaced with icon component? 🤔

assets/js/theme/global/menu.js
```diff
@@ -75,6 +82,27 @@ class Menu {
     onDocumentClick() {
         this.collapseAll()
     }
+
+    updateBodyMargin() {
+        if (this.$header.classList.contains('c-header__fixed')) {
```
aleksandraworhacz: how about renaming `c-header__fixed` to `is-fixed`? I'd avoid binding js to a components' classes :)

assets/scss/components/_wishlist-list.scss
```diff
@@ -31,3 +31,12 @@
     gap: 12px;
     align-items: center;
 }
+
+.c-wishlist__icon {
+    width: rem-calc(24px);
+    height: rem-calc(24px);
+
+    use {
+        fill: transparent;
```
aleksandraworhacz: no need of nesting `use` here, fill could be defined in parent selector

assets/js/theme/global/menu.js
mgbelegu: This was actually some code that was not needed anymore so I went ahead and deleted it :)

aleksandraworhacz (approved):
approved (however still a little sad about injecting HTML like this )

PR #150, opened 2024-04-30
XCORNER-44 product options should change image

mgbelegu:
- On product option change, fetch variants and their images, filter the selected options and return the correct image to replace with on the main image element


assets/js/theme/common/product-details.js
```diff
@@ -286,8 +290,8 @@ export default class ProductDetails extends ProductDetailsBase {
             Object.fromEntries(new FormData($form)),
             'products/bulk-discount-rates',
             (err, response) => {
-                const productAttributesData = response.data || {}
-                const productAttributesContent = response.content || {}
+                const productAttributesData = response?.data || {}
```
aleksandraworhacz: Hmm,maybe add condition like 
```
if (!response) { return; }
```

instead of checking for `response` at multiple lines?

assets/js/theme/custom/update-product-image-based-on-options.js
```diff
@@ -0,0 +1,95 @@
...
+            const optionId = getOptionId(input.name)
+            if (optionId) {
+                selectedOptions[optionId] = input.value
+            }
+        })
+        return selectedOptions
+    }
+
+    function applyImageUpdate(variants) {
+        const selectedOptions = getSelectedOptions()
```
aleksandraworhacz, line 32: Too much happens in this function. Let's create a separate function called `normalizeVariantsOptions` that will take a data and return some nice object . Also, correct me if I'm wrong, but... I think each option's input has its own unique ID, so you may store just an array of selected values (`[118,116]`) instead of having more complex object with option group ID as a key? 

So if you could store just ids of selected inputs, the whole comparing process gets so much easier, because then you can have:

```
const variants = {
  [variantID] : {
     defaultImage: [default image url]
      options: [ single_option_ID, ... ]
    }, 
  (...)
}
```

and simply find a match by ```variants.find((variant) => selectedOptions.every((option) => variant.options.includes(option)))```

Much simpler and cleaner :)

aleksandraworhacz (commented):
Change proposed :)

mgbelegu:
@aleksandraworhacz Can you please review again?

assets/js/theme/common/product-details-base.js
```diff
@@ -276,7 +276,7 @@ export default class ProductDetailsBase {
         } else {
             /* eslint-disable no-unused-expressions */
             viewModel.stock.$container?.classList.add('u-hidden-visually')
-            viewModel.stock.$input.textConent = data.stock
+            if (viewModel.stock.$input) viewModel.stock.$input.textConent = data.stock
```
aleksandraworhacz: Wrap condition's body with `{}` :)

aleksandraworhacz (approved):
Looks good ✨

PR #152, opened 2024-05-03
XCORNER-43 product card second image feature

mgbelegu:


assets/js/theme/custom/show-secondary-product-card-image.js
```diff
@@ -0,0 +1,64 @@
...
+        const secondImage = image.querySelector(selectors.secondaryImage)
+        if (secondImage) {
+            firstImage.classList.toggle('is-hidden')
+            secondImage.classList.toggle('is-hidden')
+        }
+    }
+
+    const figureWrappers = document.querySelectorAll(selectors.figureWrapper)
+    if (figureWrappers.length > 0) {
+        figureWrappers.forEach((figure) => {
```
aleksandraworhacz: You don't need to check if `figureWrappers.length` is greater than 0 if you are iterating with forEach

assets/scss/generic/_reset.scss
```diff
@@ -67,3 +67,7 @@ fieldset {
     border: 0;
     min-width: 0; /* [1] */
 }
+
+#main-content {
```
aleksandraworhacz: Don't use id for styling, and use HTML tag instead

aleksandraworhacz (approved):
Added few small comments, but overall approved :)

PR #153, opened 2024-05-13
XCORNER-51 side cart (cart preview component)

mgbelegu:


assets/js/theme/global/cart-preview.js
```diff
@@ -53,17 +54,20 @@ export default function (secureBaseUrl, cartId) {
...
 
-        if ($cart.hasClass('js-cart-disable-preview')) {
+        if ($cart.classList.contains('js-cart-disable-preview')) {
             event.preventDefault()
 
-            $cartDropdown.classList.add(loadingClass)
-            $cartDropdown.append($cartLoading)
-            $cartLoading.style.display = 'block'
+            if ($overlay) {
+                $overlay.style.display = 'block'
```
aleksandraworhacz: Why not toggling "is-hidden" classname instead?

assets/js/theme/global/cart-preview.js
```diff
@@ -8,9 +8,10 @@ export const CartPreviewEvents = {
 }
 
 export default function (secureBaseUrl, cartId) {
-    const loadingClass = 'is-loading'
+    const loaded = 'loaded'
```
aleksandraworhacz: It isn't descriptive enough. Based on just the name and value, I would have no idea what this is. So, how about creating an object called `classnames` e.g.

` const classnames = {
    loaded: "is-loaded"
}`

(more handy solution if you are planning to use more classnames like that)

or

simply rename it to `loadedClassname`. 

Also, is "loaded" a good word here? It doesn't seem like anything is loading here :D

assets/js/theme/global/cart-preview.js
```diff
@@ -53,17 +54,20 @@ export default function (secureBaseUrl, cartId) {
             return event.stopPropagation()
         }
 
-        if ($cart.hasClass('js-cart-disable-preview')) {
+        if ($cart.classList.contains('js-cart-disable-preview')) {
```
aleksandraworhacz: What's the reason for this? If `$cart` is a jQuery element (which its name suggests) the `hasClass` should work?

assets/scss/components/_overlay.scss
```diff
@@ -0,0 +1,12 @@
+.c-overlay {
+    display: none;
```
aleksandraworhacz: use `is-hidden` to toggle overlay's visibility 

e.g.

`.c-overlay {
   ...

   &.is-hidden {
     display: none;
    }
  }
```

Another issue I'm seeing here, if you toggle `display`, there would be no smooth transition or animation. Generally we want to animate element's hiding/showing to have nicer user experience :)

File: assets/scss/components/_overlay.scss [hunk 5]
```diff
@@ -0,0 +1,12 @@
...
+    content: '';
+    background-color: rgba(39, 39, 39, 0.32);
+    backdrop-filter: blur(6px);
+    top: 0;
+    left: 0;
+    position: absolute;
+    width: 100vw;
+    height: 100%;
+    z-index: $z-index-high;
```
aleksandraworhacz: I recommend to have a separate z-index variable for overlays - make it much clearer on what the variable relates to :)

assets/scss/components/_preview-cart.scss
```diff
@@ -0,0 +1,92 @@
...
+    top: 0;
+    max-width: rem-calc(390px);
+    width: 100%;
+    height: 100vh;
+    background-color: $color-white;
+    z-index: $z-index-very-high;
+    padding: rem-calc(24px);
+    padding-bottom: 0;
+
+    &.loaded {
```
aleksandraworhacz: By default, the element should be styled as shown, but you can hide it via the `is-hidden` classname. You may feel like "what's the difference?" but having a consistent approach through projects will make your life (and any developers in a future) easier :)

assets/scss/components/_preview-cart.scss
```diff
@@ -0,0 +1,92 @@
...
+    padding-bottom: rem-calc(24px);
+    margin-bottom: rem-calc(24px);
+    border-bottom: 1px solid $color-secondary-75;
+}
+
+.c-cart-preview__item-information {
+    display: flex;
+    gap: rem-calc(24px);
+
+    .c-cart__item-image-wrapper {
```
aleksandraworhacz: 1. Why nesting selectors? Couldn't you create modifier classname for item-image-wrapper? 
2. `max-width: 68px` at line 43, shouldn't be in rem units?

assets/scss/components/_preview-cart.scss
```diff
@@ -0,0 +1,92 @@
...
+
+.c-cart-preview__subtotal-container {
+    display: flex;
+    flex-direction: column;
+    gap: rem-calc(24px);
+    background-color: $color-secondary-50;
+    margin-inline: rem-calc(-24px);
+    padding: rem-calc(24px);
+
+    .c-button {
```
aleksandraworhacz: Please don't nest selectors epecially when you end up mixing different components like that. Create a modifier class for `.c-button` instead

e.g. `.c-button--subtotal` or more generic `.c-button--septenary` or sth like that

assets/scss/components/_preview-cart.scss
```diff
@@ -0,0 +1,92 @@
...
+    justify-content: flex-end;
+    align-items: center;
+    gap: rem-calc(24px);
+}
+
+.c-preview-cart__icon {
+    fill: transparent;
+}
+
+.c-cart-preview__subtotal-label,
```
aleksandraworhacz: Seems like you could have one class `.c-cart-preview__subtotal` (and prevent duplicating the same styles for two different classes) and just add `.u-color-black` / `.u-color-neutrals-300` to adjust color 🤔 

or have something like 

```
.c-cart-preview__subtotal {
   ...

   &--black { ... }
   &--grey { ... } 
}
```

templates/components/common/cart-preview.html
```diff
@@ -1,100 +1,159 @@
...
-                                    {{name}}
-                                </a>
+                        <div class="c-cart__item-head">
+                            <a class="o-text o-text--xl o-text--bold c-cart__item-title" href="{{url}}">
+                                {{name}}
+                            </a>
+                            <div class="c-cart__item-rating-wrapper">
+                                <div class="c-cart__item-rating" role="img" aria-label="">
+                                    <span class="c-cart-item__filled-stars">
+                                        {{> components/common/star status='filled' class="u-hidden"}}
```
aleksandraworhacz: It looks like a lot of repetitive code. Also, shouldn't the star icon remain the same regardless of its full/empty state and just change its filling? 

Considering the logic, there's no way the product could have 3.5 rating?

templates/components/common/cart-preview.html
```diff
@@ -1,100 +1,159 @@
...
+                                    --
+                                {{/or}}
+                                {{#or can_modify (if type '==' 'GiftCertificate')}}
+                                    <button 
+                                        class="js-cart-remove c-cart__button-remove"
+                                        data-cart-itemid="{{id}}"
+                                        data-confirm-delete="{{lang 'cart.confirm_delete'}}"
+                                        aria-label="{{lang 'cart.remove_item' name=name}}"
+                                    >
+                                        <svg class="c-cart__icon" fill="white"><use xlink:href="#icon-trash" /></svg>
```
aleksandraworhacz: icon component maybe?

aleksandraworhacz (commented):
Added comments

assets/js/theme/global/cart-preview.js
mgbelegu: For some reason it says .hasClass() or even .addClass() is not a function :/

templates/components/common/cart-preview.html
mgbelegu: @aleksandraworhacz This was previously refactored to be like this in a previous CR. Remember I used to inject the svg HTML via JS? I know it looks repetitive and there is no way with this to achieve 3.5 rating in the current format. What do you think, how should we handle this so we can just work on it once and make it perfect without changing it again in the future? My only idea is to just do it with JavaScript, but maybe there is a better solution 🤔

aleksandraworhacz: @mgbelegu I have in mind something like:

Single star code:
```
<div class="c-rating-star">
    <svg width="12" height="12" xmlns="http://www.w3.org/2000/svg" viewBox="-1 -1 14 14" class="c-rating-star__icon">
      <defs>
        <linearGradient id="star-part-{{id}}">
          <stop offset="0%" class="star-fill" />
          <stop offset="{{fill}}" class="star-fill" />
          <stop offset="{{fill}}" stop-color="white" />
          <stop offset="100%" stop-color="white" />
        </linearGradient>
      </defs>

      <path d="m6 0 1.67 4.22 4.33.37-3.3 2.97 1.01 4.44-3.71-2.38-3.71 2.38 1.01-4.44-3.3-2.98 4.33-.37z" fill="url(#star-part-{{id}})" class="c-rating-star__stroke" stroke-width="0.8" />
    </svg>
</div>
```
where `id` is a loop index and fill represents how much % of the star is supposed to be filled. 

and the `c-rating`component would be like:
```
<div class="c-rating">
    {{#for 1 5 }}
        [ still = check if value - $index > 1 -> if so set 100% else multiply value by 100% ]
        {{> components/rating/rating-star id="{{$index}}" fill="{{starFill}}%"}}
    {{/for}}
</div>

```
where the value is overall rating e.g. `3.5`


the styles would be smth like:
```
.c-rating-star {
  --icon-color: #c1a45a;

  display: inline-flex;
  font-size: 12px;
}

.c-rating-star__icon {
  width: 1em;
  height: 1em;
}

.star-fill {
  stop-color: var(--icon-color, #000);
}

.c-rating-star__stroke {
  stroke: var(--icon-color, #000);
}

```

I'm not sure tho if it's possible tho, I remember that math operations and variables in bigCommerce were really terrible in terms of what can you do :/ 

Or you could always use js to set proper `offset` values :) 

(sorry if the message was chaotic 🙈 )

mgbelegu: I've changed it. Can you please review? I think it works much better now.

Please note that I've done this on purpose:

`<div class="c-rating">
    {{> components/common/star index='1'}}
    {{> components/common/star index='2'}}
    {{> components/common/star index='3'}}
    {{> components/common/star index='4'}}
    {{> components/common/star index='5'}}
</div>`

because using the for loop for some reason it doesn't work. It gives me a partial issue saying the component cannot be found. If I add `{{#partial "page"}}` to the component the cart preview view doesn't work.

assets/js/theme/global/cart-preview.js
```diff
@@ -1,23 +1,126 @@
...
+            if (response?.data.status === 'succeed') {
+                const remove = quantity === 0
+                updateCartContent()
+                this.refreshContent(remove)
+            }
+        })
+    }
+
+    function removeCartItem(cartItemId) {
+        utils.api.cart.itemRemove(cartItemId, (err, response) => {
```
aleksandraworhacz: You could call `updateCartItem(cartItemId, 0)` without need of adding another function :)

assets/js/theme/global/cart-preview.js
```diff
@@ -1,23 +1,126 @@
...
+            if (response?.data.status === 'succeed') {
+                updateCartContent()
+                this.refreshContent(true)
+            }
+        })
+    }
+
+    function handleCartUpdateClick(e) {
```
aleksandraworhacz, line 55: I'd rename to `handleQuantityChange` - seems more accurate than "update"

assets/js/theme/global/cart-preview.js
```diff
@@ -1,23 +1,126 @@
...
+    }
+
+    function addCartUpdateFunctions() {
+        const $cartQuantityButton = q$$('.js-cart-update')
+        const $deleteItemButton = q$$('.js-cart-remove')
+        const $quantityInput = q$$('.js-cart-item-qty-input')
+
+        $cartQuantityButton?.forEach((button) => {
+            button.removeEventListener('click', handleCartUpdateClick)
+            button.addEventListener('click', handleCartUpdateClick)
```
aleksandraworhacz: Not a fan of constantly removing and adding the same listener. Why can't we determine if the listener has been already added, and if so just not adding it again?

assets/js/theme/global/cart-preview.js
aleksandraworhacz: Maybe it's not a jquery element?

assets/scss/components/_button.scss
```diff
@@ -87,6 +87,13 @@
             transition: all 0.2s ease-in-out;
         }
     }
+
+    &--with-spacing {
```
aleksandraworhacz: try to use adjectives for modifier's part

assets/scss/components/_cart.scss
```diff
@@ -55,6 +55,12 @@
     @include bp(small) {
         width: 100%;
     }
+
+    &--cart-preview {
```
aleksandraworhacz: Try to not use very specific "placement" as a modifier part of classname. It should remain more generic like "--sided"

templates/components/common/rating.html
```diff
@@ -0,0 +1,7 @@
+<div class="c-rating">
```
aleksandraworhacz, line 1: Please add comment about why this is not in loop - after months someone could change it not knowing that it was on purpose

templates/components/common/rating.html
```diff
@@ -0,0 +1,7 @@
...
+    {{> components/common/star index='1'}}
+    {{> components/common/star index='2'}}
+    {{> components/common/star index='3'}}
+    {{> components/common/star index='4'}}
+    {{> components/common/star index='5'}}
```
aleksandraworhacz, line 7: Also, add information about the rating for screen readers. It could be a span visually hidden with the text `x.x out of 5 stars`

assets/js/theme/custom/add-reviews-to-cart-items.js
```diff
@@ -1,6 +1,23 @@
...
+function setStarRating(rating) {
+    const stars = document.querySelectorAll('.c-star')
+    stars.forEach((star, index) => {
+        const fillValue = Math.max(0, Math.min(100, (rating - index) * 100))
+        const gradientId = `star-part-${index + 1}`
+        const gradient = star.querySelector(`#${gradientId}`)
+        if (gradient) {
+            const stops = gradient.querySelectorAll('stop')
+            if (stops.length === 4) {
+                stops[1].setAttribute('offset', `${fillValue}%`)
```
aleksandraworhacz: Instead of reffering to `stop` per index, you can add `js-fill` class to elements with `c-star-fill` : `<stop offset="0%" class="c-star-fill" />` and set the offset for each of them

mgbelegu:

assets/scss/components/_spinner.scss
```diff
@@ -0,0 +1,19 @@
+.c-spinner {
+    width: 40px;
+    padding: 6px;
+    aspect-ratio: 1;
+    border-radius: 50%;
+    background: $color-neutrals-300;
+    --_m: conic-gradient(#0000 10%, #000), linear-gradient(#000 0 0) content-box;
```
aleksandraworhacz: Please use more descriptive name for variable

assets/js/theme/custom/add-reviews-to-cart-items.js
```diff
@@ -1,6 +1,22 @@
...
+function setStarRating(rating) {
+    const stars = document.querySelectorAll('.c-star')
+    stars.forEach((star, index) => {
+        const fillValue = Math.max(0, Math.min(100, (rating - index) * 100))
+        const gradientId = `star-part-${index + 1}`
+        const gradient = star.querySelector(`#${gradientId}`)
+        if (gradient) {
+            const stops = gradient.querySelectorAll('stop')
+            stops.forEach((stop) => {
+                stop.setAttribute('offset', `${fillValue}%`)
```
aleksandraworhacz, line 19: Even tho the code works, it's not how the `stop` should be set according to documentation. However you could simply a code and have only two `stops` element
```
  <stop offset="50%" class="js-fill c-star-fill"></stop>
  <stop offset="0" stop-color="white"></stop>
```

where the 1st one determines how much of a space needs to be filled, and the 2nd one is always set to 0; 

Also, take a look at `fillValue` sometimes it gives weird values e.g. for rating 4.6 the last star has filling set to `59.999999999999964%`

assets/js/theme/custom/add-reviews-to-cart-items.js
```diff
@@ -1,6 +1,22 @@
 /**
  * Add reviews to cart items via GraphQL
  */
+
+function setStarRating(rating) {
+    const stars = document.querySelectorAll('.c-star')
```
aleksandraworhacz: Try not to bind js function with component's class - use js class instead :)

aleksandraworhacz (approved):
Few small comments added, but generally looks good :)

assets/js/theme/custom/add-reviews-to-cart-items.js
mgbelegu: Can you please review my updates again? I am not sure if I follow your comments regarding the `stop`.

PR #154, opened 2024-06-03
XCORNER-52 bug fixes

mgbelegu:
- Remove arrows on product card thumbnails if there is no second image
- Add some margin to the modals on mobile - 20px
- "reviews" link didn't appear in one product in side cart - **fixed**
- Add to cart modal, add some styling for product options (example: Size small, Color blue)


assets/scss/components/_preview.scss
```diff
@@ -0,0 +1,11 @@
+.c-preview__options {
+    display: flex;
+    gap: rem-calc(12px);
+}
+
+.c-preview__label {
+    &--bold {
```
aleksandraworhacz: This could be `u-bold` instead :)

aleksandraworhacz (approved):
Added one teeny tiny comment, but overall approved :)

